PR #154, opened 2015-01-03
Fix build for current rust master

kazimuth:
[Ops were removed from the prelude](https://github.com/rust-lang/rust/commit/56290a004493a5d2e211f056601533253497df60), [`Sized` is no longer implied for `Self`](https://github.com/rust-lang/rfcs/pull/546), and [the fixed orphan check](https://github.com/nikomatsakis/rust/commit/c61a0092bc) seems to have broken `#[derive(RustcEncodable, RustcDecodable)` somehow. Added op imports, sized bounds, and the `old_orphan_check` feature to fix these problems.

### DIFF
--- a/src/aabb.rs
+++ b/src/aabb.rs
@@ -28,7 +28,7 @@ use num::{zero, one, BaseNum, BaseFloat};
 use std::fmt;
 use std::num::Float;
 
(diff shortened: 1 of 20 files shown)


bjz:
Closed by #153

PR #158, opened 2015-01-09
Fix int to isize

csherratt:


bjz:
Having some issues? :P

csherratt:
I should really not be using github as a glorified diff tool :P

bjz:
:D

PR #170, opened 2015-02-11
Rust version up

simnalamburt:
Fixes compile error. `#[plugin] extern crate rand_macros;` is deprecated now.


src/cgmath.rs
```diff
@@ -33,9 +34,7 @@
 
 extern crate "rustc-serialize" as rustc_serialize;
 extern crate rand;
-#[plugin]
-#[no_link]
-extern crate rand_macros;
+#[no_link] extern crate rand_macros;
```
tomaka: I don't think you need this line.

simnalamburt: Wow you're right

PR #174, opened 2015-02-19
Remove dependencies to rand and rand_macros

tomaka:
`rand_macros` won't be available for Rust 1.0 and seems to break twice a week. I propose to just remove these trait implementations.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,5 +22,3 @@ name = "cgmath"
 
 [dependencies]
 rustc-serialize="*"
-rand_macros="*"
-rand="*"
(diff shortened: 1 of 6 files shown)


tomaka:
Oh, benches are using the traits.

aepsil0n:
For `nalgebra` I opted to implement `Rand` manually and to depend on `rand`. I think this is pretty useful functionality for some testing setups and the `rand` crate itself should be stable. Of course, it is a bit more work.

bjz:
Let's implement them manually.

csherratt:
Merged with #188

PR #176, opened 2015-02-21
Add a phantom type annotation on Transform.

bfops:
This fixes the build for me on `rustc 1.0.0-nightly (522d09dfe 2015-02-19) (built 2015-02-20)`.


kvark:
Wow, I didn't think it has to be so complex.

bfops:
It's because `S` is used to constrain the `Point` and `Vector`, but not directly as part of any of the functions. Maybe some small refactoring could let us remove `S` from the trait and have it be a constraint in the `impl` instead?

kvark:
@bfops sounds good!

PR #187, opened 2015-03-13
Generic 3D bounds

### LINKED ISSUE #117: Implement frustum intersections for Aabb and Sphere

### LINKED ISSUE #125: Frustum::from_matrix4 is wrong
For example, 
```
let persp = cgmath::PerspectiveFov {
    fovy: cgmath::Deg { s: 90.0f32 },
    aspect: 1.0,
    near: 0.1,
    far: 1000.0
};
let frustum = Frustum::from_matrix4(persp.to_matrix4());

println!("{}", frustum.far);

println!("{}", persp.to_matrix4().mul_v(&Vector4::new(0.0, 0.0, 1.0, 1.0)));
println!("{}", persp.to_matrix4().mul_v(&Vector4::new(0.0, 0.0, -1.0, 1.0)));
println!("{}", persp.to_matrix4().mul_v(&Vector4::new(0.0, 0.0, -999.0, 1.0)));
println!("{}", persp.to_matrix4().mul_v(&Vector4::new(0.0, 0.0, -1001.0, 1.0)));
```
outputs `0x + 0y + 0.000999z - 0.999999 = 0` for the far plane, i.e. z=1000, when in fact the plane should be z=-1000.

I think this is because Plane::from_vector4 uses the vector's fourth component as D, but the way from_matrix4 uses it, it should be -D. For example, the left clipping plane should have the equation v . (row(3) + row(0)) = 0 (taking the rows from the projection matrix), but since planes are represented with equations of the form A*x + B*y + C*z - D = 0, the D is negated.

I'm not sure whether Plane::from_vector4 is doing what's intended, but either it needs to negate D, or Frustum::from_matrix4 needs to be fixed somehow.

kvark:
Closes #117 
Fixes #125 

Infrastructure to do frustum culling based on `Sphere` or `Aabb` bounds. More can be implemented for the `Bound` trait. Example code:
```rust
let projection = PerspectiveFov {
        fovy: rad(1f32),
        aspect: 1f32,
        near: 1f32,
        far: 10f32,
};
let frustum = projection.to_frustum();
let bound = Sphere {
    center: Point3::new(0f32, 3f32, -5f32),
    radius: 1f32,
};
match frustum.contains(&bound) {
    Relation::In => show(),
    Relation::Cross => show_anyway(),
    Relation::Out => cull(),
}
```


tests/aabb.rs
```diff
@@ -8,7 +8,7 @@
 //     http://www.apache.org/licenses/LICENSE-2.0
 //
 // Unless required by applicable law or agreed to in writing, software
-// distributed under the License is distributed on an "AS IS" BASIS,
+// distributed under the License is distributed on an "AS IS" BASisize,
```
bjz: oopsie

kvark: nice find ;) fixed now!

tests/aabb.rs
```diff
@@ -19,53 +19,54 @@ use cgmath::{Aabb, Aabb2, Aabb3};
...
 #[test]
-fn test_aabb() {
-    let aabb = Aabb2::new(Point2::new(-20is, 30is), Point2::new(10is, -10is));
-    assert_eq!(aabb.min(), &Point2::new(-20is, -10is));
-    assert_eq!(aabb.max(), &Point2::new(10is, 30is));
-    assert_eq!(aabb.dim(), Vector2::new(30is, 40is));
-    assert_eq!(aabb.volume(), 30is * 40is);
-    assert_eq!(aabb.center(), Point2::new(-5is, 10is));
+fn test_general() {
+    let aabb = Aabb2::new(Point2::new(-20isize, 30isize), Point2::new(10isize, -10isize));
```
csherratt: Thanks for cleaning this up.

src/frustum.rs
```diff
@@ -48,14 +49,38 @@ Frustum<S> {
...
+
+    /// Find the spatial relation of a bound inside this frustum.
+    pub fn contains<B: Bound<S>>(&self, bound: &B) -> Relation {
+        [&self.left, &self.right, &self.top, &self.bottom, &self.near, &self.far]
+            .iter().fold(Relation::In, |cur, p| {
+            let r = bound.relate_plane(p);
+            // if we see Cross, then the result is Cross
+            // if we see In, then we keep the old result
+            // otherwise, take the current result
+            if cur == Relation::Cross || r == Relation::In {
```
csherratt: I think there is a bug here, if `r` is `Relation::Cross` and `cur` is `In` the state will be `Relation::In` instead of `Relation::Cross`


This might be nicer as a match too:
```rust
match (cur, r) {
    (Relation::Out, _) | (_,  Relation::Out) => Relation::Out,
    (Relation::Cross, _) | (_,  Relation::Cross) => Relation::Cross,
    _ => Relation::In
}
```

src/point.rs
```diff
@@ -445,3 +448,26 @@ impl<S: BaseNum> fmt::Debug for Point3<S> {
...
+            Relation::Out
+        }else {
+            Relation::Cross
+        }
+    }
+
+    fn relate_clip_space(&self, projection: &Matrix4<S>) -> Relation {
+        use std::cmp::Ordering::*;
+        let p = projection.mul_v(&self.to_homogeneous());
+        match (p.x.abs().partial_cmp(&p.w), p.y.abs().partial_cmp(&p.w), p.z.abs().partial_cmp(&p.w)) {
```
csherratt: Should we handle the `NaN` case?

kvark: It would just return `Cross` for the matter of `Relation`. Not sure what would be a better way to handle it.

src/frustum.rs
kvark: The condition `cur == Cross || r == In` will be false, so the outcome will take the other arm, which is just `r`, which is exactly `Cross`. So I think condition is still correct?

csherratt: You are right. What about about the case where `cur == Cross && r == Out`?

csherratt:
This needs to be rebased.

src/frustum.rs
kvark: `cur == Cross` would let it go the first arm, which results in `cur`, which is `Cross`. I believe this is still correct?

kvark:
Rebased now.

kvark:
We'd want to implement `relate_clip_space` for `Aabb3` in an efficient way, described here:
https://fgiesen.wordpress.com/2010/10/17/view-frustum-culling/

But that could be done later on without breaking the interface.

src/frustum.rs
csherratt: I think there are two cases missing here. If `cur == Out` and `r == In` The result should still be `Out` where it will be marked as `In`. The other case if `cur == Cross && r == Out` the result should be once again `Out` where it would be marked as `Cross`.

csherratt:
I'm not sure why Travis is not building this...

src/frustum.rs
kvark: `cur == Out && r == In` would result in `Out`, since it's the first arm of the condition. You are correct about the second case though, `Cross` shouldn't overwrite `Out`, and I need to fix the condition.

kvark: I fixed the condition now. Thank you for spotting the logical error. Your attention to detail is highly professional!

kvark:
It was a bad merge on my part. The upcoming build should be fine though.

kvark:
Another thing to do would be adding a benchmark for frustum culling. This would be handy for the optimized `relate_clip_space`, so will probably be implemented with it.

csherratt:
Thanks~

Issue #175, opened 2015-02-19
Implement rand traits manually


bjz:
The `rand_macros` crate breaks all the time :(

/cc. #174

PR #188, opened 2015-03-15
Manually implment Rand.

csherratt:
This rolls up #174 and completes #175. It will also should allow for #187 to build.


kvark:
Thanks for clearing the way!

PR #193, opened 2015-03-18
Various little changes.

neuschaefer:
Some mostly independent minor fixes.

### DIFF
--- a/src/line.rs
+++ b/src/line.rs
@@ -1,5 +1,5 @@
 // Copyright 2013-2014 The CGMath Developers. For a full listing of the authors,
-// refer to the AUTHORS file at the top-level directionectory of this distribution.
+// refer to the Cargo.toml file at the top-level directionectory of this distribution.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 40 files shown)


src/line.rs
```diff
@@ -38,56 +38,54 @@ impl<S: BaseNum, V: Vector<S>, P: Point<S, V>>  Line<P> {
...
+        let u = cross_2 / cross_1;
 
-                return None;
-            }
+        if zero <= t && u >= zero && u <= One::one() {
+            return Some(Point2::new(p.x + t*r.x, p.y + t*r.y));
         }
+
+        return None;
     }
```
kvark: Is this the final return? Should be just `None` then

kvark:
This is super cool. Thank you @neuschaefer !
If you are going to answer/fix my small nitpick, I'll wait. Otherwise, I'd be happy to merge.

src/line.rs
neuschaefer: Turns out to be, yes. I'll change it in a follow-up commit.

neuschaefer:
I'll fix it later (unless you're faster), feel free to merge!

kvark:
Thanks a bunch!

PR #196, opened 2015-03-26
Fixed Index implementations

kvark:


kvark:
All green. Merging as harmless.

PR #198, opened 2015-03-29
Remove old_impl_check

csherratt:
This is a _huge_ kludge to limp a little further with this library before someone actually migrates it to Associated types.


kvark:
Actually, nice to see you being able to patch it with so little blood lost ;)

bjz:
Thanks!

tomaka:
Could one of you please publish this on crates.io?

PR #201, opened 2015-04-04
Fix compilation of Vector::from_value tests

nicholasbishop:


bjz:
Seems to have been fixed!

Issue #22, opened 2014-01-10
Overload basic operators


kvark:
Vector +- Vector
Point +- Vector
Matrix * Vector
Quat * Vector
...
probably a lot more, but we need to keep the behavior strictly obvious (i.e. *not* overloading Vector*Vector as a dot product)

kvark:
I believe the trait reform is done already? Unblocking.

bjz:
I believe this is done!

PR #207, opened 2015-04-19
Allow writing `matrix * scalar`

tomaka:


bjz:
Thanks! We should probably get rid of the `.mul_s` methods now.

kvark:
Related to #22 
@bjz sure!

PR #211, opened 2015-04-25
Fixed the num crate

kvark:


tomaka:
Note that there's a `rust-1.0` branch. The version on crates.io comes from that one.

kvark:
@tomaka master still needs to be fixed. Merging myself since no one cares.

PR #214, opened 2015-05-06
Replace most To* traits by Into

tomaka:
[breaking-change]

Instead of having 10 different traits (`ToQuaternion`, `ToMatrix4`, `ToRad`, etc.), use a single one: `Into`. The `Into` trait is in the prelude.

The next step would be to modify some definitions to take a `Into<F>` instead of directly a `F`.
The from_fixed/into_fixed functions would also need to be converted.


kvark:
:+1:

tomaka:
Don't merge. Once #215 is merged, I'll rebase this one and resubmit it.

tomaka:
Closing for #216

PR #216, opened 2015-05-06
Replace most To* traits by Into

tomaka:
See #214

### DIFF
--- a/src/angle.rs
+++ b/src/angle.rs
@@ -40,38 +40,18 @@ pub struct Deg<S> { pub s: S }
 /// Create a new angle, in degrees
 #[inline] pub fn deg<S: BaseFloat>(s: S) -> Deg<S> { Deg { s: s } }
 
-/// Represents types that can be converted to radians.
-pub trait ToRad<S: BaseFloat> {
-    /// Convert this value to radians.
-    fn to_rad(&self) -> Rad<S>;
-}
-
-/// Represents types that can be converted to degrees.
-pub trait ToDeg<S: BaseFloat> {
-    /// Convert this value to degrees.
-    fn to_deg(&self) -> Deg<S>;
-}
-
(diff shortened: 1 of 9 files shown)


kvark:
I'm good with it. Calling for @cmr @bjz @csherratt to look as well.

csherratt:
I don't have any large objections, just one question is there a way [to do this](https://github.com/bjz/cgmath-rs/pull/216/files#diff-c86e2260fdda80ceb2e711a17aa630baR28) without the the extra assignment?

```rust
// maybe?
angle.into::<Deg<_>>(); 
```

Just don't want to break the ability to write one liners.

tomaka:
I think you can write `<Deg<_> as From<_>>::from(angle)`, and [type ascription](https://github.com/rust-lang/rust/issues/23416) will eventually make it cleaner.

But the major idea is that for example functions that take a `Rad<S>` should take a `T` where `T: Into<Rad<S>>` instead, so that they can handle both degrees and radians.

For example if we add `impl<S, T> Add<T> for Vector3<S> where T: Into<Vector3<S>> { ... }`, then we'll be able to write: `let v = Vector3::new(0.0, 0.0, 0.0); v + (1.0, 0.0, 0.0);`

csherratt:
Freaken awesome tomaka!

So `Vector4::new(...) * 2.0` could exist.

tomaka:
In fact you can just write `Deg::from(angle)` and it should work.

kvark:
I guess @csherratt is good with the change. Merging now.
Thanks @tomaka !

tomaka:
Next version to be published should have its minor version bumped!

FredrikNoren:
Looks like this never made it into the docs (docs still refer to ToMatrix4 etc.)

Awesome lib btw! Looks very well written

PR #217, opened 2015-05-06
Add from_rescale functions

tomaka:
The lack of these functions has always bothered me in cgmath.


kvark:
Questions:
  1. why `from_rescale` and not `from_scale`?
  2. why not accepting a vector?

tomaka:
1 - I'm bad at naming things. I was thinking `from_rescaling_factors` or something, but shorter.
2 - I don't really think it makes sense to build a `Vector` that contains the factor by which to rescale. It would have to be a separate type dedicated to this, which is probably overkill.

bjz:
Wait, don't we have `MatrixN::from_diagonal`?

PR #218, opened 2015-05-08
Make functions in angle.rs more generic

tomaka:
- Allows writing `rad(1) + deg(45)`.
- Allows using `cos`/`sin`/etc. with degrees as well.


kvark:
Nice! Ready to merge?

tomaka:
Yes

PR #219, opened 2015-05-10
Version bump

kvark:


tomaka:
Noooooo...

> Next version to be published should have its minor version bumped!

Why does nobody care about stability and semver except me?

kvark:
@tomaka ah, sorry, forgot about this note.

You could have just bumped it yourself with the breaking change, so that whoever needs to publish it doesn't need to be aware of your intentions ;)

Also, I'm a little concerned of how semver works out from the practical standpoint, especially in an environment of constant language changes. Major changes, which do not necessarily break the interface, can get minor bumps and practically remain unnoticed as far as the version goes.

PR #222, opened 2015-05-13
Implemented from_fixed() on Matrix2/3/4

Esption:
It required adding Copy to the type, but BaseFloat is only implemented for f32 and f64, which are both Copy-able, so it shouldn't affect anything.  Also, Copy is used in other parts, so again, shouldn't affect anything.

Wasn't able to use the match statement, since it's marked as experimental, but just doing a normal index access works well enough.

### DIFF
--- a/src/matrix.rs
+++ b/src/matrix.rs
@@ -557,7 +557,7 @@ impl<S: BaseFloat> One for Matrix4<S> {
     #[inline] fn one() -> Matrix4<S> { Matrix4::identity() }
 }
 
-impl<S> FixedArray<[[S; 2]; 2]> for Matrix2<S> {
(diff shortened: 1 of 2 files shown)


tomaka:
Thanks!

What about adding `+ Copy` to the `BaseFloat` trait instead?

Esption:
I didn't think about that, but sounds like a better idea.

kvark:
Why did `Copy` bound need to be replaced by `BaseFloat`?

Esption:
@kvark It gave a lot of errors when I added ```+ Copy``` to ```BaseFloat``` and changing the ```Copy``` on the ```impl<S: Copy> FixedArray``` over to ```impl<S: BaseFloat> FixedArray``` as requested by @tomaka

PR #229, opened 2015-09-20
Use travis-cargo for documentation updates

### LINKED ISSUE #223: outdated docs??
Hi there, I don't know if this is still going or not? but apparently ```let mat4 = rota.to_matrix4();``` or something like that is now ```let mat4: Matrix4<_> = rota.into();```.

Have been trying to convert the step "A simple transformation" in https://open.gl/transformations to cgmath, will continue trying but if you could provide the "exact" same translation pointed there will love it.

Also... if this will be not more under development should I switch to nalgebra or some like that? (what about aabb and so on???

bjz:
Fixes #223

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,8 @@ sudo: false
 language: rust
 
 rust:
-  - nightly
+  # - nightly
+  - beta
   - stable
 
 env:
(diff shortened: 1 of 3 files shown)


bjz:
Woops - probably shouldn't have been a minor version bump for the toml update. Sorry! >_<

PR #232, opened 2015-09-28
Move some methods onto the Matrix trait

bjz:
Now we can call them like `Matrix4::identity()`! Yay!

This also adds `Matrix::from_diagonal`, which probably closes #217.


csherratt:
LGTM!

PR #236, opened 2015-09-30
Implement binary operators for points, vectors, quaternions, and matrices

bjz:
We can't yet remove the ugly operator methods, due to rust-lang/rust#20671

This also removes the implementations of `Zero` and `One` for vectors, and removes the by-value operator implementations.


bjz:
I'm going to merge this, but please post an issue if you are unhappy with having to use the operators by-ref. We still have time before I publish the next major version.

PR #240, opened 2015-10-03
Clean up some type signatures by using associated types

bjz:


bjz:
I just realised that `Point`'s ssociated `Vector` will have to be used in the operator constraints in the future, so using an associated type would make the compiler complain about a cyclic type reference.

bjz:
This is basically what we are doing:

```rust
use std::ops::*;

trait VectorSpace {}

trait AffineSpace where
    Self: Add<Self::Diff, Output = Self>,
    Self: Sub<Self, Output = Self::Diff>,
{
    type Diff: VectorSpace;
}
```

Compiler says no, however:

```
<anon>:6:15: 6:25 error: unsupported cyclic reference between types/traits detected [E0391]
<anon>:6     Self: Add<Self::Diff, Output = Self>,
                       ^~~~~~~~~~
<anon>:6:15: 6:25 help: see the detailed explanation for E0391
note: the cycle begins when computing the bounds for type parameter `Self`...
note: ...which then again requires computing the bounds for type parameter `Self`, completing the cycle.
<anon>:7:30: 7:40 error: unsupported cyclic reference between types/traits detected [E0391]
<anon>:7     Self: Sub<Self, Output = Self::Diff>,
                                      ^~~~~~~~~~
<anon>:7:30: 7:40 help: see the detailed explanation for E0391
note: the cycle begins when computing the bounds for type parameter `Self`...
note: ...which then again requires computing the bounds for type parameter `Self`, completing the cycle.
<anon>:6:15: 6:25 error: unsupported cyclic reference between types/traits detected [E0391]
<anon>:6     Self: Add<Self::Diff, Output = Self>,
                       ^~~~~~~~~~
<anon>:6:15: 6:25 help: see the detailed explanation for E0391
note: the cycle begins when computing the supertraits of `AffineSpace`...
note: ...which then again requires computing the supertraits of `AffineSpace`, completing the cycle.
error: aborting due to 3 previous errors
playpen: application terminated with error code 101
```

http://is.gd/RsfZBQ

PR #246, opened 2015-11-03
Convert trait type parameters to associated types

bjz:
rust-lang/rust#24092 has caused quite a bit of ugliness, but I've tried to work around it as best I can.

### DIFF
--- a/src/transform.rs
+++ b/src/transform.rs
@@ -27,7 +27,7 @@ use vector::*;
 /// A trait representing an [affine
 /// transformation](https://en.wikipedia.org/wiki/Affine_transformation) that
 /// can be applied to points or vectors. An affine transformation is one which
-pub trait Transform<S: BaseNum, P: Point<S>>: Sized {
+pub trait Transform<P: Point>: Sized {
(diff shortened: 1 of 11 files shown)


bjz:
cc @csherratt @kvark - does this look ok?

kvark:
Could you explain a bit on why you are doing it?

src/transform.rs
```diff
@@ -27,24 +27,24 @@ use vector::*;
 /// A trait representing an [affine
 /// transformation](https://en.wikipedia.org/wiki/Affine_transformation) that
 /// can be applied to points or vectors. An affine transformation is one which
-pub trait Transform<S: BaseNum, V: Vector<S>, P: Point<S, V>>: Sized {
+pub trait Transform<P: Point>: Sized {
```
kvark: sweet

kvark:
Looked through it, looks nice.

bjz:
This part of the Rust Book is good: https://doc.rust-lang.org/book/associated-types.html

bjz:
Ok, I'm going to merge this.

PR #256, opened 2015-11-16
Replace ApproxEq trait with the one found in the approx crate

### LINKED ISSUE #255: Use approx crate instead of ApproxEq trait
I've just [published a library](https://github.com/bjz/approx) that provides an `ApproxEq` trait with ulps and relative equalities. Would be nice to switch cgmath over.

cc. #210

brendanzab:
Closes #255


tomaka:
You should definitely reexport the `ApproxEq` trait from within cgmath and treat it as if it was part of cgmath, otherwise it's a breakage waiting to come.

PR #261, opened 2015-12-06
Add matrix scale functions

### LINKED ISSUE #259: Add method for easily making homogeneous scale `Matrix4`s
I misunderstood #217. @tomaka says that he is constantly writing:

```rust
{ let mut m = Matrix4::from_value(size); m[3][3] = 1.0; m }
```

I'm not sure how useful this is for `Matrix2` and `Matrix3` though.

bjz:
Fixes #259

### DIFF
--- a/src/matrix.rs
+++ b/src/matrix.rs
@@ -109,7 +109,7 @@ impl<S> Matrix3<S> {
 }
 
 impl<S: BaseFloat> Matrix3<S> {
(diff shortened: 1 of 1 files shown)


bjz:
Ugh... I'm now wondering whether `from_scale` should just take a value, seeing as it is the common case. And have `from_nonuniform_scale` for the non-uniform case. Or some other name.

PR #284, opened 2015-12-22
Fix conversion from Matrix3 to Quaternion

DerekBurch:
I was getting some strange behavior when converting from a Basis3 into a Quaternion so I looked into the conversion implementation. I noticed the implementation differed slightly from the link in the function (http://www.cs.ucr.edu/~vbz/resources/quatut.pdf on page 8). I think I changed it to match the implementation in the pdf, but confirmation that things are matching now would be great. I also added a test which fails with the previous implementation and passes with these changes.

Let me know if anything is broken or anything you would like changed/improved.

### DIFF
--- a/tests/matrix.rs
+++ b/tests/matrix.rs
@@ -13,6 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#[macro_use]
 extern crate cgmath;
 
 use cgmath::*;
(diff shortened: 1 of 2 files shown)


tests/matrix.rs
```diff
@@ -365,3 +365,29 @@ fn test_from_angle() {
...
+    fn eq(a: Matrix3<f32>, b: Matrix3<f32>) {
+        if !a.approx_eq(&b) {
+            panic!("{:?} != {:?}", a, b);
+        }
+    }
+
+    fn test_angles(x: Rad<f32>, y: Rad<f32>, z: Rad<f32>) {
+        let basis: Basis3<f32> = Rotation3::from_euler(x, y, z);
+        let quaternion: Quaternion<f32> = basis.into();
+        eq(basis.into(), quaternion.into());
```
bjz: You can use the [`approx_eq!`](https://github.com/bjz/cgmath-rs/blob/master/src/approx.rs#L64) macro for this.

tests/matrix.rs
```diff
@@ -365,3 +365,29 @@ fn test_from_angle() {
...
+#[test]
+fn test_matrix3_into_quaternion()
+{
+    fn eq(a: Matrix3<f32>, b: Matrix3<f32>) {
+        if !a.approx_eq(&b) {
+            panic!("{:?} != {:?}", a, b);
+        }
+    }
+
+    fn test_angles(x: Rad<f32>, y: Rad<f32>, z: Rad<f32>) {
```
bjz: Could you make this a macro so that we can get line numbers if the comparison falis?

tests/matrix.rs
```diff
@@ -365,3 +365,29 @@ fn test_from_angle() {
...
+
+    // trigger each of the cases in: from(mat: Matrix3<S>) -> Quaternion<S>
+    // trigger the case: trace >= S::zero()
+    test_angles(rad(0.0f32), rad(0.0), rad(0.0f32));
+    // trigger the case: (mat[0][0] > mat[1][1]) && (mat[0][0] > mat[2][2])
+    test_angles(rad(2.0f32), rad(1.0), rad(-1.2f32));
+    // trigger the case: (mat[0][0] > mat[1][1]) && (mat[0][0] > mat[2][2])
+    test_angles(rad(2.0f32), rad(1.0), rad(3.0f32));
+    // trigger the else case
+    test_angles(rad(1.0f32), rad(1.0), rad(3.0f32));
```
bjz: Would be nice if these were separate tests so that we get more fine-grained failures.

bjz:
Thanks! Would be nice to get this in before I publish `0.7.0` tonight.

DerekBurch:
I pushed some changes which split the test into 4 small tests and switched to using the `assert_approx_eq!` macro. Though I didn't convert my helper function (now named `test_matrix3_into_quaternion_for_angles`) to a macro. Since the test is now split into 4 tests which each call the function once, it is probably easier to track where the failure occurred. If you still think I should convert it into a macro I can look into doing it, but I have no experience writing macros so It may take me a little bit to get right.

bjz:
Great stuff, thanks!

PR #293, opened 2016-01-01
Support for scalar on the lhs of arithmetic operators

bitshifter:
This change adds support for scalars on the left hand side of arithmetic operators, so far just for VectorN types.

In cgmath you could write code like ```Vector3::unit_y() * 2.0``` but not ```2.0 * Vector3::unit_y()```. This change adds support for the latter.

To do this, I had to add trait support for each built in numeric type. I've done this using macros, trying to keep the style close to what was already there. I noticed for the Vector operators there are also operators that take a reference, I'm not sure if I need to provide those or not?

I've also added a few tests for operators. I've used a macro to cut down on duplicate code, the down side being when it fails it's a pain to work out which line failed.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -106,26 +106,14 @@ macro_rules! impl_operator {
                 let ($lhs, $rhs) = (self, other); $body
             }
         }
-    };
-}
 
-macro_rules! impl_scalar_ops {
-    ($VectorN:ident<$S:ident> { $($field:ident),+ }) => {
-        impl_operator!(Add<$VectorN<$S>> for $S {
-            fn add(scalar, vector) -> $VectorN<$S> { $VectorN::new($(scalar + vector.$field),+) }
-        });
-        impl_operator!(Sub<$VectorN<$S>> for $S {
-            fn sub(scalar, vector) -> $VectorN<$S> { $VectorN::new($(scalar - vector.$field),+) }
-        });
-        impl_operator!(Mul<$VectorN<$S>> for $S {
-            fn mul(scalar, vector) -> $VectorN<$S> { $VectorN::new($(scalar * vector.$field),+) }
-        });
-        impl_operator!(Div<$VectorN<$S>> for $S {
-            fn div(scalar, vector) -> $VectorN<$S> { $VectorN::new($(scalar / vector.$field),+) }
-        });
-        impl_operator!(Rem<$VectorN<$S>> for $S {
-            fn rem(scalar, vector) -> $VectorN<$S> { $VectorN::new($(scalar % vector.$field),+) }
-        });
+        impl<'a> $Op<&'a $Rhs<$S>> for $Lhs {
+            type Output = $Output;
+            #[inline]
+            fn $op(self, other: &'a $Rhs<$S>) -> $Output {
+                let ($lhs, $rhs) = (self, other); $body
+            }
+        }
     };
 }
 
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -95,6 +95,38 @@ macro_rules! impl_operator {
             }
         }
     };
+    // When the left operand is a scalar
+    ($Op:ident<$Rhs:ident<$S:ident>> for $Lhs:ty {
+        fn $op:ident($lhs:ident, $rhs:ident) -> $Output:ty { $body:expr }
+    }) => {
+        impl $Op<$Rhs<$S>> for $Lhs {
+            type Output = $Output;
+            #[inline]
+            fn $op(self, other: $Rhs<$S>) -> $Output {
+                let ($lhs, $rhs) = (self, other); $body
+            }
+        }
+    };
+}
+
+macro_rules! impl_scalar_ops {
+    ($VectorN:ident<$S:ident> { $($field:ident),+ }) => {
+        impl_operator!(Add<$VectorN<$S>> for $S {
+            fn add(scalar, vector) -> $VectorN<$S> { $VectorN::new($(scalar + vector.$field),+) }
+        });
+        impl_operator!(Sub<$VectorN<$S>> for $S {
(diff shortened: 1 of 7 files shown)


src/vector.rs
```diff
@@ -285,10 +286,20 @@ macro_rules! impl_vector {
     }
 }
 
+macro_rules! impl_vector_scalar_ops {
+    ($($S:ident)*) => ($(
+        impl_scalar_ops!(Vector2<$S> { x, y });
+        impl_scalar_ops!(Vector3<$S> { x, y, z });
+        impl_scalar_ops!(Vector4<$S> { x, y, z, w });
+    )*)
+}
```
bjz: I would prefer this to be moved into the `impl_vector!` macro, to be consistent with the other codegen. So you would have:
```rust
impl_scalar_ops!($VectorN<usize> { $($field),+ });
impl_scalar_ops!($VectorN<u8> { $($field),+ });
impl_scalar_ops!($VectorN<u16> { $($field),+ });
impl_scalar_ops!($VectorN<u32> { $($field),+ });
impl_scalar_ops!($VectorN<u64> { $($field),+ });
impl_scalar_ops!($VectorN<isize> { $($field),+ });
impl_scalar_ops!($VectorN<i8> { $($field),+ });
impl_scalar_ops!($VectorN<i16> { $($field),+ });
impl_scalar_ops!($VectorN<i32> { $($field),+ });
impl_scalar_ops!($VectorN<i64> { $($field),+ });
impl_scalar_ops!($VectorN<f32> { $($field),+ });
impl_scalar_ops!($VectorN<f64> { $($field),+ });
```
I don't mind the duplication of code there.

src/macros.rs
```diff
@@ -95,6 +95,38 @@ macro_rules! impl_operator {
...
+        impl_operator!(Mul<$VectorN<$S>> for $S {
+            fn mul(scalar, vector) -> $VectorN<$S> { $VectorN::new($(scalar * vector.$field),+) }
+        });
+        impl_operator!(Div<$VectorN<$S>> for $S {
+            fn div(scalar, vector) -> $VectorN<$S> { $VectorN::new($(scalar / vector.$field),+) }
+        });
+        impl_operator!(Rem<$VectorN<$S>> for $S {
+            fn rem(scalar, vector) -> $VectorN<$S> { $VectorN::new($(scalar % vector.$field),+) }
+        });
+    };
```
bjz: Could we move this to the `vector` module?

src/macros.rs
```diff
@@ -95,6 +95,38 @@ macro_rules! impl_operator {
...
+    ($Op:ident<$Rhs:ident<$S:ident>> for $Lhs:ty {
+        fn $op:ident($lhs:ident, $rhs:ident) -> $Output:ty { $body:expr }
+    }) => {
+        impl $Op<$Rhs<$S>> for $Lhs {
+            type Output = $Output;
+            #[inline]
+            fn $op(self, other: $Rhs<$S>) -> $Output {
+                let ($lhs, $rhs) = (self, other); $body
+            }
+        }
```
bjz: Yes, it would be nice to have an impl where `$Rhs` is taken by-ref.

bjz:
This is great! Is it also possible to add impls for points, matrices and quaternions?

bitshifter:
I will look into the other types, they should be pretty similar.

tests/vector.rs
```diff
@@ -33,6 +33,109 @@ fn test_from_value() {
     assert_eq!(Vector4::from_value(76.5f64), Vector4::new(76.5f64, 76.5f64, 76.5f64, 76.5f64));
 }
 
+macro_rules! test_op_add {
+    ($VectorN:ident { $($field:ident),+ }, $s:expr, $v:expr) => (
+        // vector + vector ops
+        assert_eq!($v + $v, $VectorN::new($($v.$field + $v.$field),+));
```
bjz: I wonder if this is just re-testing the implementation? Dunno if it matters though...

bjz: Once we establish that the by-val implementation works, I'm guessing we could rely on that for the further tests. ie. `assert_eq!(&$v + &$v, $v + $v);`.

bitshifter: > I wonder if this is just re-testing the implementation? Dunno if it matters though...

Could you clarify, do you mean versus an alternative of comparing with a know good result, or something else? If it's any help, I did catch some copy and paste errors I made in some of the operator implementations with this test :)

bjz: I mean that the same method is being used to both generate the implementation, and to generate the tests. It's good to have them there for redundancy, but it would be nice to figure out some other testing method. It's not going to hold up me merging though.

src/macros.rs
bitshifter: I'm considering moving this back to ```macros``` so it can also be used by ```point```. Or would you prefer ```point``` to have it's own copy?

tests/vector.rs
bitshifter: The actual code paths are different though, it seems trivial but I think it checks assumptions are correct.

src/macros.rs
bitshifter: Actually, I take that back. ```point``` currently has no ```Add<S>``` or ```Sub<S>```, so the interface is slightly different to ```vector``` (although perhaps it should have these operators?).

bitshifter: Getting slightly offtopic, but I Point operators look a little inconsistent, perhaps this is correct and I'm not understanding it properly though:

The ```Point``` trait looks like:

```rust
    Self: Add<<Self as Point>::Vector, Output = Self>,
    Self: Sub<Self, Output = <Self as Point>::Vector>,
```

What's the logic around ```Add``` taking a ``vector`` and returning a ```point``` and ```Sub``` doing the opposite?

bjz:
:+1:

PR #321, opened 2016-04-09
Implement From<Matrix4> for AffineMatrix3

mhintz:
Enables easier interop between the structs


bjz:
Would allow clients break the invariants for a homogeneous transformation matrix? Supposedly this interface was meant to prevent that...

mhintz:
Ahhh, I see. In that case, there should be more mechanisms for building and modifying this struct. At the moment, I think the use cases are pretty limited by the interface

PR #322, opened 2016-04-09
Add a note to the readme about the column vector convention used by cgmath 

mhintz:
Adds a note to the readme to help newcomers


bjz:
Awesome! :+1: :+1:. Could you separate this out from the other commit?

PR #334, opened 2016-04-20
Fix assignment operators for angles

Arvamer:
Now they are doing something :)


bjz:
HA! woopsie

bjz:
(This is why I need more tests.... >_<)

Issue #337, opened 2016-04-22
Is there a reason why Transform3 isn't implemented for Matrix4?


mhintz:
Currently, as far as I know the only way to transform a Point3 with a Matrix4 is:

```rust
Pt::from_vec((transform * pt.to_homogeneous()).truncate())
```

But there's a perfectly good API for objects which can transform points, `Transform3`.

This would be useful for procedural geometry generation applications, where you want to produce vertices for the same shape, translated and rotated to your desired location.

kvark:
Related - https://github.com/bjz/cgmath/pull/302#issuecomment-211168768

bjz:
Agreed.

bjz:
The original reason, as mentioned in the thread that @kvark linked to, was that it requires additional interpretation of a matrix to treat it as an affine transformation vs a rotation. I think that I need to revise this stance though, for simplicity.

bjz:
Related: #324

bjz:
I'm guessing we would have to incorporate `One` (and `Zero` for completeness) into the trait hierarchy to avoid collisions. `concat` would then become `mul`.

mhintz:
Stuff like this hurts the usability of cgmath for me. I just want to be able to make quaternions, matrices, and vectors, and have them all play nice with each other in the expected ways.

bjz:
Closing in favor of #349

Issue #324, opened 2016-04-11
Add rotation constructors to Matrix4


bjz:
`Matrix4` doesn't currently implement methods like `from_euler` etc...

mhintz:
I suppose, in addition to the From<Euler> impl for Matrix4, this issue needs from_angle_x, from_angle_y, from_angle_z, and from_axis_angle?

Anything I'm forgetting?

bjz:
Fixed by #348 ❤️

PR #347, opened 2016-04-27
implement Transform for Matrix3 and Matrix4

mhintz:
Pursuant to conversations in #337, #302, and #324, I've implemented Transform for these two matrices. This makes working with the matrices as fully-fledged transformations much easier.

Along the lines of https://github.com/bjz/cgmath/pull/302#issuecomment-211168768, I've also implemented Transform2 and Transform3 for Matrix3, to reflect its potential use as both a 3D rotation matrix and a 2D affine transform.

Transform3 is implemented for Matrix4 as well, and I moved the From<Euler> impls down with the rest of the From impls, in an unrelated edit which I think nonetheless improves readability of what is now a rather large file.

I'm not positive I've gotten the Transform2 implementation for Matrix3 correct, would appreciate a sanity check :D

### DIFF
--- a/src/matrix.rs
+++ b/src/matrix.rs
@@ -756,6 +756,33 @@ impl<S: BaseFloat> ApproxEq for Matrix4<S> {
     }
 }
 
+impl<S: BaseFloat> Transform<Point3<S>> for Matrix3<S> {
+  fn one() -> Matrix3<S> {
+    One::one()
+  }
+
+  fn look_at(eye: Point3<S>, center: Point3<S>, up: Vector3<S>) -> Matrix3<S> {
+    let dir = center - eye;
+    Matrix3::look_at(dir, up)
+  }
+
+  fn transform_vector(&self, vec: Vector3<S>) -> Vector3<S> {
+    self * vec
+  }
+
+  fn transform_point(&self, point: Point3<S>) -> Point3<S> {
+    Point3::from_vec(self * point.to_vec())
+  }
+
+  fn concat(&self, other: &Matrix3<S>) -> Matrix3<S> {
+    self * other
+  }
+
+  fn invert(&self) -> Option<Matrix3<S>> {
+    SquareMatrix::invert(self)
+  }
+}
+
 impl<S: BaseFloat> Transform<Point3<S>> for Matrix4<S> {
   fn one() -> Matrix4<S> {
     One::one()
@@ -781,6 +808,8 @@ impl<S: BaseFloat> Transform<Point3<S>> for Matrix4<S> {
     SquareMatrix::invert(self)
   }
 }
+
+impl<S: BaseFloat> Transform3<S> for Matrix3<S> {}
 
 impl<S: BaseFloat> Transform3<S> for Matrix4<S> {}
 
--- a/src/matrix.rs
+++ b/src/matrix.rs
@@ -778,7 +778,7 @@ impl<S: BaseFloat> Transform<Point2<S>> for Matrix3<S> {
     self * other
   }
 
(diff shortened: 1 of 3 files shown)


mhintz:
Well, two things. One: I should note that both of my implementations of look_at for Matrix3 might be a little strange, seeing as how the function takes three parameters, and I'm condensing those down to two.

Second, I see that I should have run the tests before submitting the PR! All of the test failures are due to the fact that both SquareMatrix and Transform define an invert (or invert_self) method, and those methods are now clashing. Of course, it's possible to use Universal Function Call Syntax to disambiguate (as described here https://github.com/rust-lang/rust/issues/17151), but that's silly for a method as commonplace as invert.

I think having Transform implemented for Matrix3 and Matrix4 is a *good thing*, so I would propose either changing the name of one of the inverts (to, say, `inverse_transform` on `Transform`), or removing it from the Transform trait altogether (it feels to me like it belongs more naturally in SquareMatrix)

bjz:
I agree. Sorry, I have been tackling this privately in my spare time. Also note that `one` also conflicts.

I was thinking something like this:

```

```
      Mul<Self, Output = Self>
                 |
                 v
                One
                 |
                 v
             Transform
                 |
    *------------*-----------*
    |            |           |
    v            v           v
 Rotation   UniformScale  Translation
                 |
                 v
          NonUniformScale

```

```

bjz:
Also, make sure you run `cargo test` :)

mhintz:
That trait hierarchy looks good to me, but as far as I can tell, it doesn't solve the problem of clashing implementations of `invert`? Or am I mistaken?

bjz:
Yeah, you're right. We could have a trait for orthogonal transforms that have inverses, but we would still have to avoid the naming clash. I think leaving it out for now would be easier.

mhintz:
Hm yeah I suppose leaving it out would be the easiest approach. What's the argument against changing the name to something that doesn't clash? Just duplication? I can imagine a world where not every struct that implements Transform is also a Matrix, and you might want an inverse function just for that Transform.

Then again, I can also imagine a world in which there are non-invertible transforms (not sure what this would be, but hey maybe it's possible)

bjz:
By "guaranteed to have inverses" I mean that, for example, not all matrices have a corresponding inverse value. But I see now... we are actually returning an `Option<T>` from `Transform::invert`.

mhintz:
Oh right! I wasn't even thinking about a transform that, for example, just collapses the x-axis to zero and wouldn't be invertible. I'm starting to see your point of view - I agree that invert should be taken out of Transform.

`invert` is also defined on Rotation and SquareMatrix, but the Rotation version returns a plain Matrix whereas the SquareMatrix version returns an Option. Maybe it's a good idea to remove it from those as well? Each struct would define its own version.

kvark:
I don't see how removing `invert` is gonna do any good. People working with generic transformations should still be able to invert their stuff. Renaming it is fine though, to avoid collisions.

mhintz:
In the interest of moving forward temporarily (at least until a possible trait reorganization lands), I added changes that resolve the clashing names: Transform::invert -> Transform::inverse_transform, and Transform::invert_self -> Transform::to_inverse

mhintz:
Thanks!

bjz:
Sorry for the lateness >_<

I'm wondering if we should just remove the `SquareMatrix::invert_self` and `Transform::to_inverse` methods. They don't really do much right now...

mhintz:
Yeah, I would agree with that decision. I think there's a subtlety there which could be confusing to someone coming new to the library. Generally I think it's a good idea to have most of your operations be immutable, except when there's a real performance cost when doing it that way...

PR #348, opened 2016-04-27
implement from_angle{x, y, z} and from_axis_angle for Matrix4

mhintz:
All are duplicates of the same functions for Matrix3, with extra parameters added in to make up the full 4x4.

These are in response to #324


bjz:
Looks great!

PR #351, opened 2016-05-02
Fix typo and wording in README

danpalmer:
"paramerise" -> "parameterize", and "more than useful" -> "more than enough" (which is better wording in the context I think).


bjz:
Cheers!

Issue #349, opened 2016-04-30
New Transformation API


bjz:
Ok, so I have been experimting with a new transformation API. Here is one idea:

```rust
/// A type that can be interpreted as a geometric transformation.
pub trait Transformation: One + ScalarMul where
    Self::Scalar: BaseFloat,
{
    type Vector: InnerSpace<Scalar = Self::Scalar>;
    type Point: EuclideanSpace<Scalar = Self::Scalar, Diff = Self::Vector>;

    fn transform_vector(&self, vector: Self::Vector) -> Self::Vector;
    fn transform_point(&self, point: Self::Point) -> Self::Point;
}

/// A transformation that can apply a translation.
pub trait Translation: Transformation where
    Self::Scalar: BaseFloat,
{
    fn from_translation(translation: Self::Vector) -> Self;
}

/// A transformation that can apply a uniform scale factor.
pub trait UniformScale: Transformation where
    Self::Scalar: BaseFloat,
{
    fn from_scale(factor: Self::Scalar) -> Self;
}

/// A transformation that can apply a scale factor that is different in each
/// dimension.
pub trait NonUniformScale: UniformScale where
    Self::Scalar: BaseFloat,
{
    fn from_non_uniform_scale(scale: Self::Vector) -> Self;
}

/// A transformation that can apply a rotation.
pub trait Rotation: Transformation where
    Self::Scalar: BaseFloat,
{
    type Angle: Angle<Unitless = Self::Scalar>;
}

/// A transformation that can apply a rotation in 2-dimensional space.
pub trait Rotation2d: Rotation where
    Self::Scalar: BaseFloat,
    Self: From<<Self as Rotation>::Angle>,
{}

/// A transformation that can apply a rotation in 3-dimensional space.
pub trait Rotation3d: Rotation where
    Self::Scalar: BaseFloat,
    Self: From<Euler<<Self as Rotation>::Angle>>,
{
    fn from_axis_angle(axis: Self::Vector, angle: Self::Angle) -> Self;
    fn from_angle_x(theta: Self::Angle) -> Self;
    fn from_angle_y(theta: Self::Angle) -> Self;
    fn from_angle_z(theta: Self::Angle) -> Self;
}
```

Thoughts? Any ideas for improvements?

Still need to figure out how inversion fits into this...

cc. #347 #337

kvark:
You can have a default implementation for `transform_vector`, based on the `transform_point` one.

[...]

Issue #61, opened 2014-04-14
API is confusing


brendanzab:
@cmr has expressed concerns on #rust-gamedev that the API is confusing, especially in regards to transformations. He suggested an API closer to that of [GLM](http://glm.g-truc.net). Of course that would be rather difficult seeing as Rust's overloading isn't nearly as expressive as C++'s. I am curious however as to what you guys think regarding the current API. What could be changed and improved?

cc. @ozkriff @csherratt

bitshifter:
On a related note, you might be interested in this discussion on putting vector/matrix math into the c++ standard https://groups.google.com/a/isocpp.org/forum/#!topic/sg14/5uZE4G6qFdM

Not sure how useful it is right now but the discussion is ongoing.

tomaka:
I didn't read the whole topic, but I advocated for the same thing in Rust in the past, for exactly the same reasons as the OP.

PR #353, opened 2016-05-11
Remove unused imports

mhintz:
Fixes compile warnings

### DIFF
--- a/src/angle.rs
+++ b/src/angle.rs
@@ -21,7 +21,6 @@ use std::ops::*;
 
 use rand::{Rand, Rng};
 use rand::distributions::range::SampleRange;
-use num_traits::Float;
 use num_traits::cast;
 
 use structure::*;
--- a/src/projection.rs
+++ b/src/projection.rs
@@ -13,7 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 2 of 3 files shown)


bjz:
Shall we do a release now?

mhintz:
It seems like a lot has changed since 0.9.1, I'm in favor. I've been running off of my local master branch for a while. It would be 0.10.0 right?

Mark

mhintz:
Adding a bump here, because this is the last place where a release was mentioned. Enough has changed that I think we should release 0.10.0. I'm using the master branch in my own projects, and it's a bit tricky to manage the dependencies of other libraries which depend on cgmath...

bjz:
Oh woops, sorry. I was working on trying to get the new transformation API working, but the type system kept exploding. I'm thinking I might cherry pick the most important stuff for the next release.

mhintz:
So, just to be clear, including the changes to the Transform trait would require solving #354, #355, #349, and doing something wrt #143 and #194, right?

(And, unless I'm mistaken, it would enable closing #61?)

bjz:
You got it!

Just published v0.10.0 btw.

mhintz:
Yay! \o/

Mark

PR #290, opened 2015-12-27
Quaternion fixes

bjz:
Fixes #285

- Reorder euler angles in quaternion conversions from `(yaw, pitch, roll)` to
  `(pitch, yaw, roll)` to be consistent with the ordering in matrix conversions.
- Base the `Rotation::from_angle_{x, y, z}` default implementations off
  `Rotation::from_euler` as opposed to `Rotation::from_axis_angle`
- Adds tests for `Quaternion::from_axis_angle`, and fixes the implementation

PR #359, opened 2016-05-12
Rotation fixes

### LINKED ISSUE #285: output from from_euler for quaternions does not match from_angle_* functions
Currently the x, y, z in the function `from_euler(x: Rad<S>, y: Rad<S>, z: Rad<S>)` do not correspond with the functions `from_angle_x`, `from_angle_y`, and `from_angle_z`.

For example, I would expect the following tests to succeed, but they currently fail.
```rust
#[test]
fn test_quaternion_euler_angle_x()
{
    let angle = rad(1.0);
    let q1: Quaternion<f32> = Rotation3::from_euler(angle, rad(0.0), rad(0.0));
    let q2: Quaternion<f32> = Rotation3::from_angle_x(angle);
    assert_approx_eq!(&q1, &q2);
}

#[test]
fn test_quaternion_euler_angle_y()
{
    let angle = rad(1.0);
    let q1: Quaternion<f32> = Rotation3::from_euler(rad(0.0), angle, rad(0.0));
    let q2: Quaternion<f32> = Rotation3::from_angle_y(angle);
    assert_approx_eq!(&q1, &q2);
}

#[test]
fn test_quaternion_euler_angle_z()
{
    let angle = rad(1.0);
    let q1: Quaternion<f32> = Rotation3::from_euler(rad(0.0), rad(0.0), angle);
    let q2: Quaternion<f32> = Rotation3::from_angle_z(angle);
    assert_approx_eq!(&q1, &q2);
}
```

If the x and z are swapped in the call to from_euler then this test will pass (when maybe it shouldn't?)
```rust
#[test]
fn test_quaternion_euler_angle_x_z_swap()
{
    let angle = rad(1.0);
    let q1: Quaternion<f32> = Rotation3::from_euler(rad(0.0), rad(0.0), angle);
    let q2: Quaternion<f32> = Rotation3::from_angle_x(angle);
    assert_approx_eq!(&q1, &q2);
}
```

The behavior for Matrix3 is different and all of the following tests pass
```rust
#[test]
fn test_matrix3_euler_angle_x()
{
    let angle = rad(1.0);
    let b1: Matrix3<f32> = Matrix3::from_euler(angle, rad(0.0), rad(0.0));
    let b2: Matrix3<f32> = Matrix3::from_angle_x(angle);
    assert_approx_eq!(&b1, &b2);
}

#[test]
fn test_matrix3_euler_angle_y()
{
    let angle = rad(1.0);
    let b1: Matrix3<f32> = Matrix3::from_euler(rad(0.0), angle, rad(0.0));
    let b2: Matrix3<f32> = Matrix3::from_angle_y(angle);
    assert_approx_eq!(&b1, &b2);
}

#[test]
fn test_matrix3_euler_angle_z()
{
    let angle = rad(1.0);
    let b1: Matrix3<f32> = Matrix3::from_euler(rad(0.0), rad(0.0), angle);
    let b2: Matrix3<f32> = Matrix3::from_angle_z(angle);
    assert_approx_eq!(&b1, &b2);
}
```

jminer:
Fixes #285 and maybe #143.

I just started learning 3D graphics and ran into #285, but since I have some past experience with Euler angles and quaternions, I tried fixing it.

I started by adding tests. These tests used to fail and now pass:

- rotate_from_euler::test_x
- rotate_from_euler::test_z
- rotate_from_euler::test_y_then_z
- matrix3::rotate_from_euler::test_x_then_y
- matrix3::rotate_from_euler::test_y_then_z

I'm pretty sure the equations in #290 are wrong because Euclidean Space uses YXZ Euler angles, but cgmath is using XYZ. Reordering the angles does not give the right answer (and since axis angle to quaternion conversion currently works correctly, the hack to fix axis angles is incorrect and an indication the other equations are wrong). This NASA document gives equations for all Euler angle sequences:

http://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/19770024290.pdf

The equations that took the most work were the Euler angles to quaternion. I had to derive those equations because they aren't in the document, and Euclidean Space and Wikipedia have them in a different order than XYZ. Airplanes usually use ZYX, so that's what Wikipedia's equations are for. I'm fairly confident all the equations are correct, but I would appreciate any double checking.

What do you think of removing from the documentation any mention of roll, pitch, and yaw? Users can orient their axes however they want and even use a left-handed or right-handed coordinate system and cgmath will work fine.

If they are using it with a game engine with different axes, pitch, yaw, and roll may not correspond to x, y, and z, and calling them that in the docs could be confusing. For example, I'm using X right, Y away, and Z up, the same as Blender and CryEngine, because I think it makes more sense (but it means yaw is Z, not Y).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Changed
+
+- Fix quaternion to Euler angles and Euler angles to quaternion conversion. The axes are now
+  correct, and the order the angles are applied is XYZ.
+
 ## [v0.9.1] - 2016-04-20
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
-- Fix quaternion to Euler angles and Euler angles to quaternion conversion. The axes are now
-  correct, and the order the angles are applied is XYZ.
+- Fix Euler angles to quaternion and quaternion to Euler angles conversion. The axes are now
+  correct, and the order the angles are applied is XYZ. The conversion now matches the conversion
+  from axis angle.
+- Fix Euler angles to matrix conversion.
 
 ## [v0.9.1] - 2016-04-20
 
--- a/src/euler.rs
+++ b/src/euler.rs
@@ -27,6 +27,17 @@ use num::BaseFloat;
 ///
 /// This type is marked as `#[repr(C, packed)]`.
 ///
+/// The axis rotation sequence is XYZ. That is, the rotation is first around
+/// the X axis, then the Y axis, and lastly the Z axis (using intrinsic
+/// rotations). Since all three rotation axes are used, the angles are
+/// Tait–Bryan angles rather than proper Euler angles.
+///
+/// # Ranges
+///
+/// - x: [-pi, pi]
+/// - y: [-pi/2, pi/2]
+/// - z: [-pi, pi]
+///
 /// # Defining rotations using Euler angles
 ///
 /// Note that while [Euler angles] are intuitive to define, they are prone to
@@ -42,9 +53,9 @@ use num::BaseFloat;
 ///
 /// For example, to define a quaternion that applies the following:
 ///
-/// 1. a 45° rotation around the _x_ axis
-/// 2. a 180° rotation around the _y_ axis
-/// 3. a -30° rotation around the _z_ axis
+/// 1. a 90° rotation around the _x_ axis
+/// 2. a 45° rotation around the _y_ axis
+/// 3. a 15° rotation around the _z_ axis
 ///
 /// you can use the following code:
 ///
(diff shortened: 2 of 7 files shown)


bjz:
Sorry for the late reply. This looks really good. Would you be able to help look over this with me @mhintz?

bjz:
It's good to see the tests! Perhaps using the vector axis methods might be clearer in the tests? Like `Vector3::unit_x()`. Also, we are missing tests for `Matrix4`. In the future I think it might be wise to abstract those tests into macros, to make sure the ones for quaternions and matrices stay in sync.

bjz:
Re. roll-pitch-yaw, @nbennet seemed to be in favour of it: https://github.com/bjz/cgmath/issues/329#issuecomment-211121190 - would be good to get their opinion based on what you mentioned (to me it sounds reasonable).

mhintz:
Thanks a lot for the commit, and especially for the work that went into it! To be honest, I don't feel like I understand quaternion math well enough to give a confident check of this work. The tests certainly help, and I trust @jminer's work. I'm in favor of merging, and closing the relevant issues.

I agree also that wherever we use the terms "roll", "pitch", or "yaw" in the code, we should include a clarifying note in the documentation. I was under the impression that those names were standard, guess I was wrong.

Thanks Jordan!

bjz:
The names are standard, it's more about what names map to what axes.

PR #360, opened 2016-05-16
Adds nightly serde support

### LINKED ISSUE #313: Serde serialization support
As a Rust-newbie coming from C# I've found `RustcSerializable` and `RustcDeserializable` are very hard to use (seems like I'm prone to [this](http://stackoverflow.com/questions/27650643/how-do-i-fix-encodable-decodable-deprecation) error), my program does very much depend on serialization though.

Switching to Serde fixed my compiler problems and also seems to be the de-facto new way of doing serialization in Rust according to [this](https://users.rust-lang.org/t/whats-up-with-rustc-serialize/589) and [this](https://www.reddit.com/r/rust/comments/3v4ktz/differences_between_serde_and_rustc_serialize/). Automatically deriving Serde implementations in my code doesn't work though, since this library is built upon `rustc-serialize`. 
So what do you think about incorporating Serde support into this library (while possibly retaining rustc-serialize support via Cargo.toml and #cfg)? I'm happy to file a PR once somebody approves this.

Liotitch:
There is a few things to note:
- I made `rustc-serialize` dependency optional to allow choosing between the two crates.

- Serde macros does not seem to support associated types. I had to implement the trait manually for `Decomposed`

- I did not use `syntex` for stable rust support as I was unsure how to handle the build configuration with both `serde_macros` and `serde_codegen` as optional dependencies.

Closes #313

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,3 +40,4 @@ serde_macros = { version = "0.7", optional = true }
 
 [dev-dependencies]
 glium = "0.14.0"
+serde_json = "0.7"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,11 +29,14 @@ name = "cgmath"
 [features]
 unstable = []
 default = ["rustc-serialize"]
+eders = ["serde", "serde_macros"]
 
 [dependencies]
 num-traits = "0.1"
 rand = "0.3"
 rustc-serialize = { version = "0.3", optional = true }
+serde = { version = "0.7", optional = true }
+serde_macros = { version = "0.7", optional = true }
 
 [dev-dependencies]
 glium = "0.14.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,11 +28,12 @@ name = "cgmath"
 
 [features]
 unstable = []
+default = ["rustc-serialize"]
 
 [dependencies]
 num-traits = "0.1"
 rand = "0.3"
-rustc-serialize = "0.3"
+rustc-serialize = { version = "0.3", optional = true }
 
 [dev-dependencies]
 glium = "0.14.0"
--- a/src/transform.rs
+++ b/src/transform.rs
@@ -148,3 +148,150 @@ impl<S: BaseFloat, R: Rotation3<S>> From<Decomposed<Vector3<S>, R>> for Matrix4<
 impl<S: BaseFloat, R: Rotation2<S>> Transform2<S> for Decomposed<Vector2<S>, R> {}
 
 impl<S: BaseFloat, R: Rotation3<S>> Transform3<S> for Decomposed<Vector3<S>, R> {}
+
+#[cfg(feature = "eders")]
+#[doc(hidden)]
+mod eders_ser {
+    use structure::VectorSpace;
+    use super::Decomposed;
+    use serde::{self, Serialize};
+
+    impl<V: VectorSpace, R> Serialize for Decomposed<V, R>
+        where V: Serialize, V::Scalar: Serialize, R: Serialize
+    {
+        fn serialize<S>(&self, serializer: &mut S) -> Result<(), S::Error>
+            where S: serde::Serializer
+        {
+            serializer.serialize_struct("Decomposed", DecomposedVisitor {
+                value: self,
+                state: 0,
+            })
+        }
+    }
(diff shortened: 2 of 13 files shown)


Liotitch:
I forgot about my debugging file sorry :cry: ! I just removed it.

bjz:
Awesome, looking now! Could you squash that debug-file deletion commit into the previous one?

Cargo.toml
```diff
@@ -28,11 +28,15 @@ name = "cgmath"
 
 [features]
 unstable = []
+default = ["rustc-serialize"]
+eders = ["serde", "serde_macros"]
```
bjz: Is `"eders"` a standard feature name in the mysterious land of serde?

src/transform.rs
```diff
@@ -147,3 +148,150 @@ impl<S: BaseFloat, R: Rotation3<S>> From<Decomposed<Vector3<S>, R>> for Matrix4<
...
+    
+    impl<V: VectorSpace, R> Serialize for Decomposed<V, R>
+        where V: Serialize, V::Scalar: Serialize, R: Serialize
+    {
+        fn serialize<S>(&self, serializer: &mut S) -> Result<(), S::Error>
+            where S: serde::Serializer
+        {
+            serializer.serialize_struct("Decomposed", DecomposedVisitor {
+                value: self,
+                state: 0
```
bjz: Missing trailing comma here.

src/transform.rs
```diff
@@ -147,3 +148,150 @@ impl<S: BaseFloat, R: Rotation3<S>> From<Decomposed<Vector3<S>, R>> for Matrix4<
...
+            })            
+        }
+    }
+    
+    struct DecomposedVisitor<'a, V: 'a + VectorSpace, R: 'a> {
+        value: &'a Decomposed<V, R>,
+        state: u8
```
bjz: And here

src/transform.rs
```diff
@@ -147,3 +148,150 @@ impl<S: BaseFloat, R: Rotation3<S>> From<Decomposed<Vector3<S>, R>> for Matrix4<
...
+                    self.state += 1;
+                    Ok(Some(try!(serializer.serialize_struct_elt("rot", &self.value.rot))))
+                },
+                2 => {
+                    self.state += 1;
+                    Ok(Some(try!(serializer.serialize_struct_elt("disp", &self.value.disp))))
+                },
+                _ => {
+                    Ok(None)
+                }
```
bjz: And here

src/transform.rs
```diff
@@ -147,3 +148,150 @@ impl<S: BaseFloat, R: Rotation3<S>> From<Decomposed<Vector3<S>, R>> for Matrix4<
...
+mod eders_de {
+    use structure::VectorSpace;
+    use super::Decomposed;
+    use serde::{self, Deserialize};
+    use std::marker::PhantomData;
+    
+    enum DecomposedField {
+        Scale,
+        Rot,
+        Disp
```
bjz: And here

src/transform.rs
```diff
@@ -147,3 +148,150 @@ impl<S: BaseFloat, R: Rotation3<S>> From<Decomposed<Vector3<S>, R>> for Matrix4<
...
+                Some(disp) => disp,
+                None => try!(visitor.missing_field("disp")),
+            };
+            
+            try!(visitor.end());
+            
+            Ok(Decomposed { scale: scale, rot: rot, disp: disp })
+        }
+    } 
+}
```
bjz: No newline hat the end of the file

bjz: Also ensure you have your editor set to strip out the whitespace - seems not to have been done.

bjz:
Could you add the feature to the .travis.yml when building the tests on nightly? Not sure how to do that... it would also be nice to have some tests for the custom serializers/deserializers, to test that they are idempotent when composed together.

Liotitch:
I have chosen `eders` similar to what the `vec-map` crate did chose. 
For the other changes i will work them out as soon as i can. Thank you for your comments !

Liotitch:
Should I implement `ApproxEq` for `Decomposed` or directly compare each fields for the test ?

bjz:
Woo, looks good!

PR #366, opened 2016-07-26
Fix 'Documentation' link in README.

c0gent:
I left the '/index.html' but I suppose it'd be better without...


brendanzab:
Fixed in #368. Thanks!

PR #367, opened 2016-07-27
Made uses of Rad<S> more generic using Into<Rad<S>>

adudney:
This normalizes the use of `Rad<S>` to the same as the `perspective` function.


brendanzab:
Thanks!

PR #371, opened 2016-08-03
decrease approx_epsilon

hmeyer:
Before computing the inverse of a 0.01 scale matrix failed, due to approx_eq resulting in the det being approx_eq to zero.


brendanzab:
Seems to make the build fail... could you look into it? Really gotta get around to getting #256 fixed and merged.

Marckvdv:
I believe this pull request is not relevant anymore, as the library has switched to using a external crate for approximate equality.

brendanzab:
Thanks for looking @Marckvdv! :)

PR #374, opened 2016-08-22
Changed over to the approx crate, updated glium and serde

adudney:
Updated https://github.com/brendanzab/cgmath/pull/256 to the latest master.

A few caveats of the current state of this PR:

* Currently breaks rotation's doc tests, so the offending lines are commented out. I'm not sure why the macros from the approx crate aren't visible in the doc test, but I couldn't find out how to make them visible.
* Macros aren't reexported, so ulps_eq! and such have to be imported from approx in the tests and in any external code that uses them. I'm not sure if reexporting macros is even supported by rust.
* Matrix.rs uses the original epsilon (1.0e-6) as most smaller values I tested caused tests to fail.

There may be other issues, this is a first pass. Let me know if I need to change anything :)


adudney:
`Aster` is failing on nightly :-/

brendanzab:
Yay, thanks!

Issue #375, opened 2016-08-23
Weird crashes on android/arm because of memory alignment


ozkriff:
See: https://github.com/ozkriff/zoc/issues/197

tomaka:
https://github.com/rust-lang/rust/issues/27060

I'm surprised that cgmath uses `#[repr(packed)]`. It definitely shouldn't.

brendanzab:
We should remove these then! Thanks for the link @tomaka. Should `#[repr(C)]` should still remain though?

tomaka:
I'd say no, but I don't have a strong opinion on that.
If you want to guarantee that the transmutes to fixed-sized arrays work, in my opinion adding tests is enough.

ozkriff:
It fixes the problem. Bump version? :)

PR #376, opened 2016-08-25
Removes #[repr(packed)] from all structs where it was specified.

mhintz:
This is intended as a partial fix for #375. It doesn't also remove #[repr(C)],
since it's a bit unclear whether that's necessary.


brendanzab:
Cheers!

brendanzab:
Sorry! was waiting for the build to pass and forgot to merge! (insert facepalm gif here)

PR #381, opened 2016-09-11
Add flip_y_up_to_z_up method to Matrix4

FredrikNoren:
This is for converting between models which often come in a y-up coordinate system (maya) and game which often use a z-up system.

Math from http://stackoverflow.com/questions/22571364/threejs-matrix-from-z-up-coordinate-system-to-y-up-coordinate-system

### DIFF
--- a/src/matrix.rs
+++ b/src/matrix.rs
@@ -299,6 +299,18 @@ impl<S: BaseFloat> Matrix4<S> {
                      S::zero(),
                      S::one())
     }
+    
+    /// Create a matrix which flips the y and z coordnates.
+    ///
+    /// This is used to transform from a y-up coordinate system to
+    /// a z-up coordinate systems.
+    #[inline]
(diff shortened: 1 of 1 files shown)


brendanzab:
Aha, interesting! Will have to think about what the best API/naming should be

FredrikNoren:
Yeah I was going back and fort a bit. Maybe something like just flip_yz is
easier

brendanzab:
How likely is it that there are other coordinate systems? And will we need this for the other transformation types? cc. @mhintz

FredrikNoren:
I think y vs z up is the most common issue.
https://www.reallusion.com/iclone/help/3dxchange5/pro/05_Export/Y_Up_or_Z_Up.htm
for instance. Generally it seems like many 3D tools (but not all) use y up
whereas game engines are split between them (unity uses y and unreal z)

PR #382, opened 2016-09-14
Version bump

brendanzab:


brendanzab:
Sorry for taking so long @ozkriff!

brendanzab:
So it seems like it is failing on the benches - everything else seems fine though.

PR #385, opened 2016-10-03
Implement component wise number casting for the Point and Matrix types.

### LINKED ISSUE #380: `cast` method for Point, Matrix, ...
I've been wondering why the [`cast` method](https://docs.rs/cgmath/0.11.0/cgmath/struct.Vector1.html#method.cast) is only implemented for `Vector*` and not for other types such as `Point*` or `Matrix*`. Is there a deeper mathematical truth I'm not understanding or what's the reason?

Marckvdv:
This fixes #380 by directly implementing these functions for these types, in the same way they were already implemented for the `Vector` type. This also adds tests to test the new features.

### DIFF
--- a/src/matrix.rs
+++ b/src/matrix.rs
@@ -14,7 +14,7 @@
 // limitations under the License.
 
 use rand::{Rand, Rng};
-use num_traits::cast;
+use num_traits::{cast, NumCast};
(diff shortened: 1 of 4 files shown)


brendanzab:
Thanks!

PR #386, opened 2016-10-08
Removed `one` function from `Transform` trait

Marckvdv:


brendanzab:
Thanks! It would be nice however for `Transform` to be constrained by `One` now, seeing as the method has been removed. I can't remember why, but I was running into problems with that...

kvark:
@Marckvdv are you planning to revive this?

PR #389, opened 2016-12-18
Use serde_derive instead of serde_macros.

bitshifter:
It looks like serde_macros was removed in serde 0.8.11 - https://github.com/serde-rs/serde/commit/1b6fd5a362ae1925153213ab329678a4ce2e2ce4.


dtolnay:
`feature(proc_macro)` is no longer needed on Rust 1.15.

PR #399, opened 2017-04-04
Bump to serde 0.9

### LINKED ISSUE #398: serde needs a version bump to 0.9
This would let one use `serde_derive` to serialize values on stable Rust. Unfortunately it's not *just* a version bump, some alteration has to happen in how the serialization actually happens.  It may be able to just derive everything, but I don't have bandwidth to dig into it right now.

Mange:
Use Serde 0.9 that no longer requires the usage of Nightly. Fixes #398.

Also adds a test run for `--features eders` to Travis config.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,6 @@ rust:
   - stable
 
 script:
-  - cargo build && cargo test
+  - cargo build && cargo test && cargo test --features "eders"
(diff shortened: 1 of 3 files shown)


Mange:
Oh, that didn't work. I had to run tests with `cargo --features eders`. I missed that part.

I'll update with the necessary fixes.

brendanzab:
Thanks!

PR #400, opened 2017-04-13
Fix between_vectors for opposite vectors

andystanton:
Quaternion::between_vectors does not work for vectors in opposite directions.

The source code references a stackoverflow answer, but there is another answer which does take into account opposite vectors: http://stackoverflow.com/a/11741520/2074937. This PR is an implementation of that answer.

### DIFF
--- a/tests/quaternion.rs
+++ b/tests/quaternion.rs
@@ -258,3 +258,57 @@ mod rotate_from_axis_angle {
         assert_ulps_eq!(vec3(-2.0f32.sqrt() / 2.0, 0.0, 2.0f32.sqrt() / 2.0), rot * vec);
     }
 }
+
+mod rotate_between_vectors {
+    use cgmath::*;
+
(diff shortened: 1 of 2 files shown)


tests/quaternion.rs
```diff
@@ -258,3 +258,57 @@ mod rotate_from_axis_angle {
...
+    #[test]
+    fn test_around_z_180_ccw() {
+        let expected = Quaternion::new(0.0, 0.0, 0.0, -1.0);
+
+        let a = vec3(-3.0, 0.0, 0.0);
+        let b = vec3(40.0, 0.0, 0.0);
+
+        assert_ulps_eq!(Quaternion::between_vectors(a, b), expected);
+    }
+}
```
brendanzab: Awesome! Missed a new line here, but seeing as there are tests I'll take it! 🎉

PR #406, opened 2017-04-24
iter traits: impl `iter::{Sum, Product}`

mathstuf:
This allows one to easily sum up a sequence of vectors or matrices.

### DIFF
--- a/tests/matrix.rs
+++ b/tests/matrix.rs
@@ -97,6 +97,18 @@ pub mod matrix2 {
     }
 
     #[test]
+    fn test_sum_matrix() {
+        let res: Matrix2<f64> = [A, B, C].iter().sum();
+        assert_eq!(res, A + B + C);
+    }
+
+    #[test]
+    fn test_product_matrix() {
+        let res: Matrix2<f64> = [A, B, C].iter().product();
+        assert_eq!(res, A * B * C);
+    }
+
+    #[test]
     fn test_determinant() {
         assert_eq!(A.determinant(), -2.0f64)
     }
@@ -256,6 +268,18 @@ pub mod matrix3 {
                                 48.0f64, 111.0f64, 174.0f64));
 
         assert_eq!(A * B, &A * &B);
+    }
+
+    #[test]
+    fn test_sum_matrix() {
+        let res: Matrix3<f64> = [A, B, C, D].iter().sum();
+        assert_eq!(res, A + B + C + D);
+    }
+
+    #[test]
+    fn test_product_matrix() {
+        let res: Matrix3<f64> = [A, B, C, D].iter().product();
+        assert_eq!(res, A * B * C * D);
     }
 
     #[test]
@@ -613,6 +637,18 @@ pub mod matrix4 {
                                 130.0f64, 306.0f64, 482.0f64, 658.0f64));
 
         assert_eq!(A * B, &A * &B);
+    }
+
+    #[test]
+    fn test_sum_matrix() {
+        let res: Matrix4<f64> = [A, B, C, D].iter().sum();
+        assert_eq!(res, A + B + C + D);
+    }
+
+    #[test]
+    fn test_product_matrix() {
+        let res: Matrix4<f64> = [A, B, C, D].iter().product();
+        assert_eq!(res, A * B * C * D);
     }
 
     #[test]
--- a/tests/vector.rs
+++ b/tests/vector.rs
@@ -20,6 +20,7 @@ extern crate cgmath;
 
 use cgmath::*;
 use std::f64;
+use std::iter;
 
 #[test]
 fn test_constructor() {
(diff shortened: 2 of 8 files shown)


brendanzab:
Nice, thanks! Would you be able to add some tests! I'm sorry, my testsuite is a bit of a mess at the moment, but it would be great to have something there for them.

brendanzab:
It might also make sense to constrain `VectorSpace` by `iter::Sum` too, seeing as it has both `Zero` and `Add` available to it.

brendanzab:
And perhaps we could add `iter::Product` to `SquareMatrix` :)

mathstuf:
I'm having issues adding `iter::Product` for `Rotation`; the trait setup is complaining about `P` not being constrained in:

```rust
impl<P: EuclideanSpace, T: Rotation<P>> iter::Product for T {
    #[inline]
    fn product<I: Iterator<Item=Self>>(iter: I) -> Self {
        iter.fold(Self::one(), Mul::mul)
    }
}
```

which I guess I understand…but I've been unable to figure out a way to write it so that it's happy either :/ .

brendanzab:
Bleh - feel free to skip `Rotation` for now, sorry for causing you pain. Those traits have been exploding weirdly for me too.

brendanzab:
Ohhh, wait. You are trying to implement it for `T`, which obviously would not work. I meant add it to the list of trait constraints on the definition of `Rotation`, then follow the type errors to see what other impls of `Product` you need to add.

mathstuf:
Oh, OK.

brendanzab:
You can leave `Rotation` for now, but I woul like to see:

```rust

pub trait VectorSpace: Copy + Clone where
    Self: Zero,

    Self: Add<Self, Output = Self>,
    Self: Sub<Self, Output = Self>,

    Self: iter::Sum,

    ...
```

And the same for `SquareMatrix`. 😄

brendanzab:
Note that `Matrix`es are also `VectorSpace`s, so you'll only need to add `iter::Sum` to the latter trait.

mathstuf:
Hmm, things aren't working out with the tests :( . Maybe it's too early in the morning, but I'll push my WIP to get the builders on it.

brendanzab:
No stress, thanks a bunch for your help tonight!

tests/matrix.rs
```diff
@@ -97,6 +99,26 @@ pub mod matrix2 {
     }
 
     #[test]
+    fn test_sum_matrix() {
+        // This doesn't work because there isn't an impl for &Matrix2<f64>.
+        // let res: Matrix2<f64> = [A, B, C].iter().sum();
+
+        // This type resolves at least, but is ugly.
+        let res: Matrix2<f64> = [A, B, C].iter().cloned().sum();
```
brendanzab: Could always do `vec![A, B, C].into_iter().sum()`... not sure if that is any better though... ugh, another allocation. It's only a test though, I guess. No need for the comment if you want to leave it though - I think it is clear enough.

tests/matrix.rs
```diff
@@ -97,6 +99,26 @@ pub mod matrix2 {
...
+                   Matrix2::new(5.0f64, 8.0f64,
+                                6.0f64, 11.0f64));
+    }
+
+    #[test]
+    fn test_product_matrix() {
+        // Can't figure out what type `P` is?
+        assert_eq!([A, B, C].into_iter().product(),
+                   Matrix2::new(42.0f64, 54.0f64,
+                                55.0f64, 67.0f64));
```
brendanzab: Wonder if the `Rotation` constraint is messing you up... perhaps try dropping that for now and seeing if it works once you make it consistent with the `test_sum_matrix` that clones the items being iterated over? Transformations need an overhaul anyway.

mathstuf: No, that isn't it. the `sum` test works because of the explicit type on `res`. I don't know why it can't figure this out automatically :/ .

tests/matrix.rs
mathstuf: Well, I'd rather not have to do the `.cloned()` bit. I wonder if it should be implemented for the reference types as well…

mathstuf: Eh, that opened a can of worms nevermind :/ .

mathstuf:
Hmm. I wonder what the expected value for `[A, B, C].product()` is for `Matrix`. Currently, you get `(A * B) * C` which probably isn't what one would expect.

mathstuf:
D'oh, mixed up commutivity and associativity. The multiplication order doesn't matter.

tests/matrix.rs
mathstuf: So I figured that out; no longer needs a `.cloned()`.

brendanzab:
Seems like you need to duplicate the vector impl into the `impl_vector_default!` macro as well - Travis is failing when simd is enabled.

tests/vector.rs
```diff
@@ -87,6 +88,15 @@ macro_rules! impl_test_rem {
     )
 }
 
+macro_rules! impl_test_iter_sum {
+    ($VectorN:ident { $($field:ident),+ }, $ty:ty, $s:expr, $v:expr) => (
+        // iter(vector).sum() ops
```
brendanzab: What is this comment?

mathstuf: Copypasta from the `div` macro; I originally though this might be more complicated :) .

brendanzab:
Thanks for your hard work, this is awesome!

mathstuf:
This likely requires a bump to `0.14` due to the additional constraints on the traits.

brendanzab:
v0.14 is now on cargo!

PR #407, opened 2017-04-26
Sum and Product cleanups

brendanzab:
This continues on from #406

I cleaned up the impls to be consistent with the other code in cgmath, and implemented `Sum` for `Angle`s. I also reordered the assertions so that type inference worked, and ensured that the tests covered both taking the iterator items by value _and_ by ref.

### DIFF
--- a/src/rotation.rs
+++ b/src/rotation.rs
@@ -159,17 +159,17 @@ impl<S: BaseFloat> From<Basis2<S>> for Matrix2<S> {
     fn from(b: Basis2<S>) -> Matrix2<S> { b.mat }
 }
 
-impl<S: BaseFloat> iter::Product for Basis2<S> {
+impl<S: BaseFloat> iter::Product<Basis2<S>> for Basis2<S> {
     #[inline]
-    fn product<I: Iterator<Item=Self>>(iter: I) -> Self {
-        iter.fold(Basis2 { mat: Matrix2::identity() }, Mul::mul)
+    fn product<I: Iterator<Item=Basis2<S>>>(iter: I) -> Basis2<S> {
+        iter.fold(Basis2::one(), Mul::mul)
     }
 }
 
-impl<'a, S: 'a + BaseFloat> iter::Product<&'a Self> for Basis2<S> {
+impl<'a, S: 'a + BaseFloat> iter::Product<&'a Basis2<S>> for Basis2<S> {
     #[inline]
-    fn product<I: Iterator<Item=&'a Self>>(iter: I) -> Self {
-        iter.fold(Basis2 { mat: Matrix2::identity() }, Mul::mul)
+    fn product<I: Iterator<Item=&'a Basis2<S>>>(iter: I) -> Basis2<S> {
+        iter.fold(Basis2::one(), Mul::mul)
     }
 }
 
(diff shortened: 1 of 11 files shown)


src/rotation.rs
```diff
@@ -159,17 +159,17 @@ impl<S: BaseFloat> From<Basis2<S>> for Matrix2<S> {
...
     }
 }
 
-impl<'a, S: 'a + BaseFloat> iter::Product<&'a Self> for Basis2<S> {
+impl<'a, S: 'a + BaseFloat> iter::Product<&'a Basis2<S>> for Basis2<S> {
     #[inline]
-    fn product<I: Iterator<Item=&'a Self>>(iter: I) -> Self {
-        iter.fold(Basis2 { mat: Matrix2::identity() }, Mul::mul)
+    fn product<I: Iterator<Item=&'a Basis2<S>>>(iter: I) -> Basis2<S> {
+        iter.fold(Basis2::one(), Mul::mul)
```
mathstuf: I don't think this is right; the `One` impl uses `MatrixN::one()` which is filled with `1`, not an identity.

src/rotation.rs
```diff
@@ -279,17 +279,17 @@ impl<S: BaseFloat> From<Basis3<S>> for Quaternion<S> {
...
     }
 }
 
-impl<'a, S: 'a + BaseFloat> iter::Product<&'a Self> for Basis3<S> {
+impl<'a, S: 'a + BaseFloat> iter::Product<&'a Basis3<S>> for Basis3<S> {
     #[inline]
-    fn product<I: Iterator<Item=&'a Self>>(iter: I) -> Self {
-        iter.fold(Basis3 { mat: Matrix3::identity() }, Mul::mul)
+    fn product<I: Iterator<Item=&'a Basis3<S>>>(iter: I) -> Basis3<S> {
+        iter.fold(Basis3::one(), Mul::mul)
```
mathstuf: Same comment as `Basis2`.

tests/vector.rs
```diff
@@ -90,9 +89,8 @@ macro_rules! impl_test_rem {
 
 macro_rules! impl_test_iter_sum {
     ($VectorN:ident { $($field:ident),+ }, $ty:ty, $s:expr, $v:expr) => (
-        let res: $VectorN<$ty> = iter::repeat($v).take($s as usize).sum();
-        assert_eq!(res,
-                   $VectorN::new($($v.$field * $s),+));
+        assert_eq!($VectorN::new($($v.$field * $s),+),
+                   iter::repeat($v).take($s as usize).sum());
```
mathstuf: Hmm. Is `assert_eq!` asymmetric in the way it does type deduction? It wouldn't figure out the type for me…

mathstuf:
I suppose tests for `Basis` product is warranted :) .

src/rotation.rs
brendanzab: MatrixN::one() should be an identity matrix... :/ - it's 'one' in the sense of being the [multiplicative identity](http://mathworld.wolfram.com/MultiplicativeIdentity.html). We only have the `SquareMatrix::identity` method because it is a bit more clear to folks - 'identity' usually refers to the multiplicative identity in the context of matrices.

tests/vector.rs
brendanzab: Well... more that it could not deduce the type parameter of the `sum` method symmetrically. The other solution would be:

```rust
assert_eq!(iter::repeat($v).take($s as usize).sum::<$VectorN<_>>(),
                  $VectorN::new($($v.$field * $s),+));
```

But I found swapping them to look more pleasing, despite not following the pattern of putting the value being tested first...

src/rotation.rs
mathstuf: Oh, `from_value` is a diagonal matrix, not a filled matrix.

PR #410, opened 2017-05-01
Workaround rust-lang/rust#41478.

kennytm:
Replace all `impl .. for Struct<<A as Angle>::Unitless>` by `Struct<A::Unitless>`. 

This allows `-Zsave-analysis` to work, and I think this is more readable too.


brendanzab:
Nice! I have the `<T as Trait>::U` stuff around in some places because I was having to work around Rust requiring me to be more explicit... maybe that has improved within that time? Would be interesting to see if any more of my annotations could be removed.

ozkriff:
@brendanzab Can you publish 0.14.1? :)

brendanzab:
Waiting for #411 to pass now! :)

brendanzab:
Oh, I remember, it was this issue: rust-lang/rust#24092

Issue #390, opened 2016-12-31
Decomposed::look_at and Matrix4::look_at produce different results


aloucks:
I noticed my view matrix wasn't quite right when using `Decomposed::look_at`. I compared the output from `Matrix4::look_at` and there were some discrepancies in the resulting matrix. After inverting, I can see that the translation appears to be correct, but the rotation is wrong. 

It looks like the error is here: https://github.com/brendanzab/cgmath/blob/master/src/transform.rs#L84

I started getting the expected results when I changed:
```rust
let rot = R::look_at(center - eye, up);
```
to 
```rust
let rot = R::look_at(eye - center, up);
```

I used this to compare:

```rust
extern crate cgmath;

use cgmath::prelude::*;
use cgmath::BaseFloat;
use cgmath::{Point3, Vector3, Basis3, Matrix4, Decomposed, Quaternion};

fn main() {
    let eye = Point3::new(1.0, 3.0, 5.0);
    let center = Point3::origin();
    let up = Vector3::unit_y();

    let trans: Matrix4<f32> = Decomposed::<Vector3<f32>, Quaternion<f32>>::look_at(eye, center, up)
        .inverse_transform().unwrap().into();
    println!("Decomposed::<Vector3<f32>, Quaternion<f32>>::look_at");
    pretty_print(trans);

    let trans: Matrix4<f32> = Decomposed::<Vector3<f32>, Basis3<f32>>::look_at(eye, center, up)
        .inverse_transform().unwrap().into();
    println!("Decomposed::<Vector3<f32>, Basis3<f32>>::look_at");
    pretty_print(trans);

    let trans: Matrix4<f32> = Matrix4::look_at(eye, center, up)
        .inverse_transform().unwrap().into();
    println!("Matrix4::<f32>::look_at");
    pretty_print(trans);

    let trans: Matrix4<f32> = look_at::<Point3<f32>, Quaternion<f32>>(eye, center, up)
        .inverse_transform().unwrap().into();
    println!("look_at::<Vector3<f32>, Quaternion<f32>> (eye - center)");
    pretty_print(trans);

    let trans: Matrix4<f32> = look_at::<Point3<f32>, Basis3<f32>>(eye, center, up)
        .inverse_transform().unwrap().into();
    println!("look_at::<Vector3<f32>, Basis3<f32>> (eye - center)");
    pretty_print(trans);
}

// https://github.com/brendanzab/cgmath/blob/master/src/transform.rs#L83
[...]

PR #412, opened 2017-05-02
Addressing `look_at` inconsistency

DaseinPhaos:
#390  pointed out the inconsistency, specifically of `Matrix4f::look_at()` among other `look_at` methods in the crate.
Reduced form:
```
let eye = Point3::new(0.0, 0.0, 0.0);
let p = Point3::new(1.0, 0.0, 0.0);
let up = vec3(0.0, 1.0, 0.0);

let pt3 = <Matrix3<_> as Transform<_>>::look_at(eye, p, up).transform_point(p);
let pt4 = Matrix4::look_at(eye, p, up).transform_point(p);

assert_relative_ne!(pt3, pt4);
assert_relative_eq!(pt3, Point3::new(0.0, 0.0, 1.0));
assert_relative_eq!(pt4, Point3::new(0.0, 0.0, -1.0));
```

This can be really confusing to the user if she hasn't been notified in advance. Therefore, this PR suggests we either
- address this inconsistency in docs more explicitly, or
- make it consistent with the rest, and introduce a new method `Matrix4::look_at_neg_z` which retains the current functionality of `Matrix4f::look_at`.

Personally I'd prefer the second approach because IMO it'd be nice if we can look at any directions we want conveniently (as `Matrix3::look_at` has to be coupled with yet another translation, to get what `Matrix4::look_at` can provide). However, it does bring breaking changes to the API.

I took the second approach in the PR, with some tests and docs.  Comments appreciated!

### DIFF
--- a/src/matrix.rs
+++ b/src/matrix.rs
@@ -93,8 +93,8 @@ impl<S: BaseFloat> Matrix2<S> {
         Matrix2 { x: c0, y: c1 }
     }
 
-    /// Create a transformation matrix that will cause a vector to point at
-    /// `dir`, using `up` for orientation.
+    /// Create a homogeneous transformation matrix that will cause a vector `dir`
+    /// to point towards `Vectors::unit_y()`, using `up` for orientation.
     pub fn look_at(dir: Vector2<S>, up: Vector2<S>) -> Matrix2<S> {
         //TODO: verify look_at 2D
         Matrix2::from_cols(up, dir).transpose()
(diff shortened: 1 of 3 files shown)


src/matrix.rs
```diff
@@ -126,8 +126,8 @@ impl<S: BaseFloat> Matrix3<S> {
         Matrix3 { x: c0, y: c1, z: c2 }
     }
 
-    /// Create a rotation matrix that will cause a vector to point at
-    /// `dir`, using `up` for orientation.
+    /// Create a homogeneous transformation matrix that will cause a vector `dir`
+    /// to point towards `Vector3::unit_z()`, using `up` for orientation.
```
brendanzab: Is this right though? This returns a 3x3 rotation matrix...?

PR #420, opened 2017-06-16
Added project_on to InnerSpace

DataAnalysisCosby:
Adds a project method to InnerSpace.
IMPORTANT NOTE: I did not check to see if this compiles because I am a bad person :(
I thought I should ask before doing that, should a test be added for this? The correctness of this method somewhat relies on dot and magnitude2 being correct.

### DIFF
--- a/src/structure.rs
+++ b/src/structure.rs
@@ -246,6 +246,15 @@ pub trait InnerSpace: VectorSpace where
     fn lerp(self, other: Self, amount: Self::Scalar) -> Self {
         self + ((other - self) * amount)
     }
+
+    /// Returns the
+    /// [vector projection](https://en.wikipedia.org/wiki/Vector_projection)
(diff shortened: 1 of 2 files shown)


brendanzab:
I think it would be nice to have some tests. Alas, I have also been bad in adding tests to cgmath, and there are various ways that I have implemented them in the library. Could you have a go at adding them anyway?

brendanzab:
Sorry for the late reply - got back from holidays this week and have been fighting jet lag!

DataAnalysisCosby:
Sure, I'll add a few tests when I have some time. And no problem with the late reply, I'm not waiting on this to go through or anything like that. Just thought it would be a useful method.

DataAnalysisCosby:
Hey brendan, just managed to subdue my laziness and add some tests. Hope they work for you!

brendanzab:
Thanks! ❤️

PR #421, opened 2017-07-13
Add cast method to Quaternion

pengowen123:
I'm not sure if the doc comment is correct (I just copied it from the vector method).


brendanzab:
Cheers!

PR #422, opened 2017-07-31
added len function to Array

hclarke:
This is to allow functions to use Index, IndexMut, etc. generically over different length vectors

possible changes:
- bikeshedding the name
- put it on its own trait


brendanzab:
Cheers!

PR #423, opened 2017-08-01
update glium to version 0.17

hclarke:
old version of cocoa wasn't building on nightly (it's a dependency of glium->glutin->cocoa and glium->glutin->winit->cocoa on osx)
updating glium fixes it


brendanzab:
Thanks!

PR #424, opened 2017-08-01
condensed duplicate code for trait specialization using macros

hclarke:
I put all the "default" keyword stuff into a impl_fn! macro. that way, there's way less duplicate code for simd trait implementations

I also feel like there should be a "specialization" or "nightly" feature that's separate from "simd" (that simd enables/depends on). This patch doesn't include that, but makes it easier to implement, since there's only one macro to change

also made a BaseSigned trait so that you can use the trait implementation macros for Neg

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -17,6 +17,53 @@
 
 #![macro_use]
 
+/// generates an impl fn, allowing specialization if simd is enabled
+macro_rules! impl_fn {
+    
+
+    // Mutable Binary
+    (fn $op:ident(&mut $lhs:ident, $rhs:ident:$Rhs:ty) $body:block ) => {
+        #[cfg(feature = "simd")]
+        #[inline]
+        default fn $op(&mut $lhs, $rhs: $Rhs) $body
+
+        #[cfg(not(feature = "simd"))]
+        #[inline]
+        fn $op(&mut $lhs, $rhs: $Rhs) $body
+    };
+
+    // Unary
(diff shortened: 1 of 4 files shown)


hclarke:
ping! @brendanzab 

have any thoughts on this one?

I mainly did it because the first PR I submitted failed on the simd travis builds the first time, and I had to fix it by duplicating code. Doing it this way means that changes to the trait implementations only need to happen in one place, so it should reduce friction on future patches

brendanzab:
Woopsie, sorry I looked, thought it was awesome, then forgot to reply!

Yeah, I re-read the simd stuff a while back and got a tad confused. In fact I think a whole lot of our macro stuff is a bit ugly, so I'm glad you've been looking into it!

src/macros.rs
```diff
@@ -17,102 +17,148 @@
 
 #![macro_use]
 
+/// generates an impl fn, allowing specialization if simd is enabled
+macro_rules! impl_fn {
```
brendanzab: Think I'd prefer this called `impl_operator_method!`

src/macros.rs
```diff
@@ -17,102 +17,148 @@
...
+    
+
```
brendanzab: Extra space here!

hclarke:
these changes are in my queue. will probably get to it in the next few days
thanks for feedback!

kvark:
@hclarke looks like the few days got stretched quite a bit :)

PR #425, opened 2017-08-09
Implement num_traits::Bounded for Points, Vectors, and Angles

Osspial:
Implements [`Bounded`](http://rust-num.github.io/num/num_traits/bounds/trait.Bounded.html) for all Points, Vectors, and Angles where `S` also implements `Bounded`, giving `min_value` and `max_value` functions.

Should `Bounded` be re-exported from the crate root? This PR currently doesn't, but other `num_traits` traits are exported from `cgmath`'s root

### DIFF
--- a/src/angle.rs
+++ b/src/angle.rs
@@ -22,7 +22,7 @@ use std::ops::*;
 
 use rand::{Rand, Rng};
 use rand::distributions::range::SampleRange;
-use num_traits::cast;
+use num_traits::{cast, Bounded};
 
 use structure::*;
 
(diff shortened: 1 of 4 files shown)


brendanzab:
Cool, thanks! Let's export `Bounded` as well. Out of curiosity, what was your use case?

Osspial:
I was using it to set initial integer values for points in a GUI layout engine, but looking at that code now it looks like it should have been initialized to zero instead! Still, it makes sense for these traits to be implemented.

PR #427, opened 2017-08-16
Modify cast functions to return Option

Osspial:
Right now, the `cast()` functions simply panic when a cast fails, crashing the entire program unhandleably (ignoring panic hooks). This PR makes said functions return `Option<_>`, so that a program may catch and handle failed casts. Unfortunately this is a breaking change, but programs should be able to handle a failed cast which can reasonably happen even within properly-implemented code.


brendanzab:
Yep, this actually makes sense.

PR #431, opened 2017-09-28
Adding inverse transformation of a vector to the Transform trait.

Rhuagh:
This is heavily used in collision detection. It's much cheaper to do the inverse transform of a ray/support function direction vector/etc, than to transform all the vertices of a primitive/complex shape.

We've found ourselves writing: `transform.inverse_transform().and_then(|inverse| Some(inverse.transform_vector(some_vec)))` over and over, so we felt there was a solid use case for having this in cgmath. It is also something that specific implementations of the Transform trait can do a much better version of than doing the full inverse of the transform, since vector space has no concept of translation.

### DIFF
--- a/src/transform.rs
+++ b/src/transform.rs
@@ -37,6 +37,11 @@ pub trait Transform<P: EuclideanSpace>: Sized {
     /// Transform a vector using this transform.
     fn transform_vector(&self, vec: P::Diff) -> P::Diff;
 
+    /// Inverse transform a vector using this transform
+    fn inverse_transform_vector(&self, vec: P::Diff) -> Option<P::Diff> {
+        self.inverse_transform().and_then(|inverse| Some(inverse.transform_vector(vec)))
+    }
+
     /// Transform a point using this transform.
     fn transform_point(&self, point: P) -> P;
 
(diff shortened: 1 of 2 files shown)


src/transform.rs
```diff
@@ -94,6 +99,15 @@ impl<P: EuclideanSpace, R: Rotation<P>> Transform<P> for Decomposed<P::Diff, R>
     }
 
     #[inline]
+    fn inverse_transform_vector(&self, vec: P::Diff) -> Option<P::Diff> {
+        if ulps_eq!(self.scale, &P::Scalar::zero()) {
```
brendanzab: Do we need an `ulps_eq!` here? We're just trying to avoid the `NaN`, right?

brendanzab: Interesting, I see we have one on `inverse_transform`. Were you going for consistency with my code? 🤔

brendanzab:
Looks good! Let me know your thoughts on the my question. If you're cool with it, I'll merge.

src/transform.rs
Rhuagh: Yes, I went with the consistency

Issue #429, opened 2017-09-03
Swizzle Operators


elrnv:
I'm wondering if there is interest in adding swizzle functions like in GLSL/GLM to vector and point (https://www.khronos.org/opengl/wiki/Data_Type_(GLSL)#Swizzling). If so, I have permutations already working in https://github.com/elrnv/cgmath, that I can make a PR for.

brendanzab:
Nice, I like these! Go ahead if you like! Sorry for the late reply btw 😓

elrnv:
Done in #432

PR #432, opened 2017-09-30
Swizzle Operators

elrnv:
Implements #429.
I extended the swizzle operators I originally proposed to a more complete set allowing repeating variables as well as making larger vectors from smaller ones. This PR effectively makes creating new vectors from old ones as simple and terse as calling `abcd()` on a point or vector where `a`,`b`,`c`, and `d` are one (or none) of the variables available in the original point or vector. For example if we have
```rust
let v = Vector3::new(1.0, 2.0, 3.0);
```
then `v.xyxz()` produces a
```rust 
Vector4 { x: 1.0, y: 2.0, z: 1.0, w: 3.0 }
```
and `v.zy()` produces a 
```rust 
Vector2 { x: 3.0, y: 2.0 }
```

Instead of hardcoding all the operators I added a build.rs script that effectively generates a `swizzle_operator_macro.rs` file in the output directory, which is then included in `macros.rs` at compile time. I think this is a more maintainable solution, but I could also alternatively commit the generated `swizzle_operator_macro.rs` directly if that is preferred.

Enjoy!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ name = "cgmath"
 
 [features]
 unstable = []
+swizzle = []
 
 [dependencies]
 approx = "0.1"
--- a/README.md
+++ b/README.md
@@ -32,6 +32,32 @@ vectors"), meaning when transforming a vector with a matrix, the matrix goes
 on the left. This is reflected in the fact that cgmath implements the
 multiplication operator for Matrix * Vector, but not Vector * Matrix.
 
+## Features
+
+### Swizzling
+This library offers an optional feature called
+["swizzling"](https://en.wikipedia.org/wiki/Swizzling_(computer_graphics))
+widely familiar to GPU programmers. To enable swizzle operators, pass the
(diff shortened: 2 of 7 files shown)


brendanzab:
Awesome work. Do you know how much this adds to our compiled binary size? Is this something we should perhaps put behind a feature flag?

elrnv:
Thanks! I agree. I think GLM also puts these guarded by an #ifdef. On my machine, adding the swizzles increases the binary size from 2.7M to 3.3M on release and to 3.4M on debug.
I added the feature guard in the last commit.

elrnv:
If you would like this enabled by default, I can change the feature to "noswizzle". The extra 0.6MB of bloat in the static library will generally be optimized away by the compiler. So AFAIK this shouldn't be an issue for the majority of users.

brendanzab:
Let's go with optional for now. Thanks for the work!

norru:
I personally like the opt-out idea better than opt-in.

PR #435, opened 2017-10-23
Add Matrix4::look_at_dir (fix #433).

### LINKED ISSUE #433: Matrix4::look_at and direction
The current interface for `Matrix4::look_at` makes it easy to look at a center point from an eye point, but harder to look at a direction vector from an eye point. To do that, one can do:

```rust
Matrix4::look_at(eye, eye + dir, up)
```

but then the `look_at` function does `(eye + dir) - eye` to retrieve the direction, which can lead to rounding errors, especially with the `f32` number type when `eye` is far from the origin and `dir` is a unit vector.

https://github.com/brendanzab/cgmath/blob/3543ac02745c43091d16560a63ca1c346c10bd1f/src/matrix.rs#L236

Would it make sense to add a `look_at_dir` function?

```rust
pub fn look_at_dir(eye: Point3<S>, dir: Vector3<S>, up: Vector3<S>) -> Matrix4<S>
```

Internally, the logic would be the same, and `look_at` could just call `look_at_dir(eye, center - eye, up)` so the change is less than 10 lines of code. Btw, the doc (mistakenly) mentions `dir` instead of `center` so the spirit is already here.

https://github.com/brendanzab/cgmath/blob/3543ac02745c43091d16560a63ca1c346c10bd1f/src/matrix.rs#L233-L235

I can propose a PR if needed.

gendx:


brendanzab:
Nice stuff! Thanks for the reasoning in #433!

PR #437, opened 2017-10-26
Use transmute_copy for into impl

Xaeroxe:
Fixes https://github.com/brendanzab/cgmath/issues/436

So while this isn't a perfect no-op like I hoped for it's still a slight improvement over what was present.  A transmute without _copy wasn't possible because even though it is possible to prove Vector3\<S\> is the same size as [S; 3] rustc just isn't that smart yet.  So this is the next best thing.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -142,7 +142,7 @@ macro_rules! fold_array {
 /// Generate array conversion implementations for a compound array type
 macro_rules! impl_fixed_array_conversions {
     ($ArrayN:ident <$S:ident> { $($field:ident : $index:expr),+ }, $n:expr) => {
-        impl<$S> Into<[$S; $n]> for $ArrayN<$S> {
+        impl<$S: Copy> Into<[$S; $n]> for $ArrayN<$S> {
             #[inline]
             fn into(self) -> [$S; $n] {
                 unsafe { mem::transmute_copy(&self) }
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -145,7 +145,7 @@ macro_rules! impl_fixed_array_conversions {
         impl<$S> Into<[$S; $n]> for $ArrayN<$S> {
             #[inline]
             fn into(self) -> [$S; $n] {
-                match self { $ArrayN { $($field),+ } => [$($field),+] }
+                unsafe { mem::transmute_copy(&self) }
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/macros.rs
```diff
@@ -163,11 +163,11 @@ macro_rules! impl_fixed_array_conversions {
             }
         }
 
-        impl<$S: Clone> From<[$S; $n]> for $ArrayN<$S> {
+        impl<$S> From<[$S; $n]> for $ArrayN<$S> {
             #[inline]
             fn from(v: [$S; $n]) -> $ArrayN<$S> {
                 // We need to use a clone here because we can't pattern match on arrays yet
```
brendanzab: Is this comment stale now? 🤔

brendanzab: Maybe we could just remove it. Good stuff! 👍

brendanzab: Wait.... is this really safe though? What happens if we pass something with a destructer in? Or custom clone impl? Exceedingly unlikely, but we might be violating the invariants of the type system here...

Xaeroxe: If $S has a Drop impl wouldn't it be called when the resulting array is dropped?  We still only have one copy of the internal data.  BaseNum also requires Copy so custom clone is out of the question.

Xaeroxe: Oh wait I need a mem::forget for the custom Drops.

Xaeroxe: Isn't the necessity of that also removed by the Copy restraint though come to think of it?

brendanzab: `Copy` types can still impl `Drop`. We probably would want to protect this impl anyway with some bounds in this case. With `T: Copy` at least. Again, this is a very unlikely case due to the fact that this will only likely ever be taking floats or ints, but it's important to be wary around `unsafe` and be good citizens of the ecosystem!

Xaeroxe: If a copy type impls drop it's inherently unsafe as the drop call might be called multiple times and treated as the same data.

Xaeroxe: I've added the restraint on the impl anyways

brendanzab:
Seem to have a build failure 🤔

Xaeroxe:
I'm not really certain this is that much of an improvement over what's already here so I'll close it

PR #438, opened 2017-11-02
Angle: add the normalize_zero method

mathstuf:
This method is like `normalize` except that it normalizes to have an
absolute value of no more than `turn_div_2`.

---
This is useful for making sure that an angle is no more than some offset from a target angle (e.g., implementing maximum turn rates in games).

### DIFF
--- a/src/structure.rs
+++ b/src/structure.rs
@@ -603,6 +603,13 @@ where
         }
     }
 
+    /// Return the angle, normalized to the range `[-turn_div_2, turn_div_2)`.
+    #[inline]
+    fn normalize_signed(self) -> Self {
+        let rem = self.normalize();
+        if Self::turn_div_2() < rem { rem - Self::full_turn() } else { rem }
+    }
+
     /// Return the angle rotated by half a turn.
     #[inline]
     fn opposite(self) -> Self {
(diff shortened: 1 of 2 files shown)


src/structure.rs
```diff
@@ -592,6 +592,13 @@ where
         }
     }
 
+    /// Return the angle, normalized to the range `[-turn_div_2, turn_div_2)`.
+    #[inline]
+    fn normalize_zero(self) -> Self {
```
kvark: everything about this PR looks good to me, but I'm not sure about the name. Would you find `normalize_signed` to be more descriptive?

Rhuagh: `normalize_to_turn_div_2()` maybe ?

mathstuf: Yeah, I'm not the biggest fan of the name as-is either. I personally like `normalize_signed` more of those options. I'll rebase once there's consensus on a name.

Rhuagh: `normalize_angle` ?

mathstuf: `normalize_for_diff` or `normalize_for_offset` (possibly without `for_` on each).

kvark: I think `normalize_signed` is most appropriate. @Rhuagh are you strongly opposed?

Rhuagh: No, I'm just brainstorming

mathstuf: Renamed to `normalize_signed`.

PR #444, opened 2018-01-05
Implement ElementWise for PointN

Osspial:
Adds element-wise functions to `Point` types. Makes sense to implement because `ElementWise` exists for mainly for pragmatic reasons, and there are cases where doing element-wise operations on `Point`s makes sense.


brendanzab:
Cheers!

PR #446, opened 2018-01-21
Add map functions to points and vectors

Osspial:
These are similar to the `map` function on the `Option` type, but perform the function on every scalar in the point or vector and put the result into a new point or vector of the same dimension. This would be useful for performing more complex element-wise operations than the ones found in the `ElementWise` trait, although unfortunately I can't see a way to implement it in that trait without having associated type constructors.

### DIFF
--- a/src/point.rs
+++ b/src/point.rs
@@ -64,33 +64,6 @@ pub struct Point3<S> {
     pub z: S,
 }
 
-impl<S: BaseNum> Point1<S> {
-    #[inline]
-    pub fn new(x: S) -> Point1<S> {
-        Point1 { x: x }
-    }
-
-    impl_swizzle_functions!(Point1, Point2, Point3, S, x);
-}
-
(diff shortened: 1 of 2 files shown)


Osspial:
Would it be possible to look over this PR soon? I have some code I'd like to put on crates.io in the next few days that depends on changes in this PR, and I'd like to point my dependencies at the crates.io version of cgmath.

spearman:
I guess you can already do this, but its ugly:

```
let mut v : cgmath::Vector3 <f64> = [1.0, 1.0, 1.0].into();
<AsMut <[f64; 3]>>::as_mut(&mut v).into_iter().for_each (foo);
```

brendanzab:
Ugh, sorry for the lateness! Merging now!

brendanzab:
Yeah, I've thought in the past about ways of implementing this generically, but never figured it out.

PR #452, opened 2018-04-21
[WIP] Implement swing/twist decomposition

richard-uk1:
This PR implements a decomposition of a quaternion into a swing and a twist.

I have realised that I don't actually need this functionality to do what I want, but since I have implemented it I thought I would submit a PR to see if there was interest in merging it. If there is, and people have valid use cases, I can finish off handling the degenerate case.

### DIFF
--- a/src/quaternion.rs
+++ b/src/quaternion.rs
@@ -93,10 +93,8 @@ impl<S: BaseFloat> Quaternion<S> {
     ///
     /// Return the closest rotation that turns `src` vector into `dst`.
     ///
-    /// - [Related StackOverflow question]
-    ///   (http://stackoverflow.com/questions/1171849/finding-quaternion-representing-the-rotation-from-one-vector-to-another)
-    /// - [Ogre implementation for normalized vectors]
-    ///   (https://bitbucket.org/sinbad/ogre/src/9db75e3ba05c/OgreMain/include/OgreVector3.h?fileviewer=file-view-default#cl-651)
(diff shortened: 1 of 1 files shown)


src/quaternion.rs
```diff
@@ -173,6 +169,40 @@ impl<S: BaseFloat> Quaternion<S> {
...
+    /// let swing = rotation * twist.conjugate();
+    /// // if we recombine our rotations we get the original back.
+    /// assert_ulps_eq!(rotation, swing * twist);
+    /// panic!("swing: {:#?}, twist: {:#?}, rotation: {:#?}, new rotation: {:#?}",
+    /// Euler::from(swing), Euler::from(twist), Euler::from(rotation), Euler::from(swing * twist));
+    /// # }
+    /// ```
+    /// . There is a degenerate case where the direction is orthogonal to the direction of the
+    /// quaternion, and the angle is 180 degrees. In this case there are infinite possible
+    /// decompositions, so allow the user to specify one with `fallback`.
```
brendanzab: I like this documentation! ❤️  - Wondering if it would be better just to return an `Option` in that case?

derekdreery: Yeah could easily do that, if it's preferable. What do you think about the utility of the whole method?

derekdreery: I think I'd like to leave this open until someone who understands this stuff has a look and says whether it's useful or not.

derekdreery:
I'm closing this since peeps seem to be coalescing around nalgebra

PR #455, opened 2018-04-28
Add is_finite method to vectors and matrices.

derekdreery:
This method is useful in debug to error early when some kind of singularity is encountered.

### DIFF
--- a/src/matrix.rs
+++ b/src/matrix.rs
@@ -106,6 +106,11 @@ impl<S: BaseFloat> Matrix2<S> {
 
         Matrix2::new(c, s, -s, c)
     }
+
+    /// Are all entries in the matrix finite.
+    pub fn is_finite(&self) -> bool {
+        self.x.is_finite() && self.y.is_finite()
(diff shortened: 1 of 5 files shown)


brendanzab:
Nice! Wonder if we could add this to the `Array` trait? I can accept this as-is though!

derekdreery:
I'll add Array, no worries! Would you like it added to anything else?

derekdreery:
I think it's difficult to add to Array, because it only makes sense on types that are `BaseFloat`. Do you agree?

brendanzab:
You can do `where Self::Element: BaseFloat` - one of the reasons I ask is that `Quaternion` is missing here - theis would have been picked up if it were on the trait.

derekdreery:
Slightly off-topic, but in Array you have swap_elements: how do you know that i and j are in bounds?

brendanzab:
I think that would just panic with an index out of bounds error. We should probably have tests for that though 😅

PR #459, opened 2018-06-08
Bumped rand to 0.5

VictorKoenders:
Rand 0.5 has a couple of warnings and breaking changes:

`Rand` is deprecated and replaced with a Distribution construction:

```rust
impl<S: BaseFloat + SampleRange> Rand for $Angle<S> {
    #[inline]
    fn rand<R: Rng>(rng: &mut R) -> $Angle<S> {
        $Angle(rng.gen_range(cast(-$hi).unwrap(), cast($hi).unwrap()))
    }
}
```
becomes
```rust
impl<S> Distribution<$Angle<S>> for Standard 
	where Standard: Distribution<S>,
		S: BaseFloat + SampleUniform {
	#[inline]
	fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> $Angle<S> {
		$Angle(rng.gen_range(cast(-$hi).unwrap(), cast($hi).unwrap()))
	}
}
```

`IsaacRng::new_unseeded` is deprecated, and rand recommends using [FromEntropy](https://docs.rs/rand/0.5.0/rand/trait.FromEntropy.html). Other options are (from rand's docs):
- Deterministic creation using SeedableRng::from_seed with a fixed seed
- Seeding from thread_rng: SeedableRng::from_rng(thread_rng())?; this will usually be faster and should also be secure, but requires trusting one extra component.

I also added some `#![allow(unused_macros)]`'s to the benchmarks to stop RLS from complaining.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ swizzle = []
 approx = "0.2"
 mint = { version = "0.5", optional = true }
 num-traits = "0.2"
-rand = "0.4"
+rand = "0.5"
 serde = { version = "1.0", features = ["serde_derive"], optional = true }
 simd = { version = "0.2", optional = true }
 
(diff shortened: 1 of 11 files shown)


brendanzab:
Thanks so much!

PR #462, opened 2018-07-08
Replace most usages of BaseFloat with a new trait BaseReal

andrewhickman:
This adds a new trait `BaseReal` which corresponds to [`num_traits::real::Real`](https://docs.rs/num-traits/0.2.5/num_traits/real/trait.Real.html) and replaces most uses of `BaseFloat`. This allows using `cgmath` with types that have no infinity or NaN values, such as fixed point arithmetic. 

I'm not sure if this is a breaking change; `BaseReal` is implemented for `T: BaseFloat` so this should strictly relax the trait bounds. I checked a couple of dependencies and they compiled fine.

### DIFF
--- a/src/angle.rs
+++ b/src/angle.rs
@@ -15,20 +15,20 @@
 
 //! Angle units for type-safe, self-documenting code.
 
-use std::fmt;
 use std::f64;
+use std::fmt;
 use std::iter;
 use std::ops::*;
 
-use rand::Rng;
-use rand::distributions::{Distribution, Standard};
-use rand::distributions::uniform::SampleUniform;
 use num_traits::{cast, Bounded};
+use rand::distributions::uniform::SampleUniform;
+use rand::distributions::{Distribution, Standard};
+use rand::Rng;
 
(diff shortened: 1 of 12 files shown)


src/angle.rs
```diff
@@ -136,42 +167,42 @@ macro_rules! impl_angle {
...
-        });
-        impl_assignment_operator!(<S: BaseFloat> MulAssign<S> for $Angle<S> {
-            fn mul_assign(&mut self, scalar) { self.0 *= scalar; }
-        });
-        impl_assignment_operator!(<S: BaseFloat> DivAssign<S> for $Angle<S> {
-            fn div_assign(&mut self, scalar) { self.0 /= scalar; }
-        });
-
-        impl<S: BaseFloat> approx::AbsDiffEq for $Angle<S> {
+        impl_operator!(<S: BaseReal> Add<$Angle<S> > for $Angle<S> {
```
Rhuagh: These seem to have gotten weird formatting, I guess rustfmt is the culprit here?

PR #463, opened 2018-07-11
Fix links to OpenGL API in docs

LukasKalbertodt:
Those links were broken before, since the [] and () part were split
into two different lines. This also caused warnings on nightly, which
are now fixed.


LukasKalbertodt:
Ping. Could someone press the merge button? :)

Osspial:
Done. Sorry that took so long to get through!

PR #469, opened 2018-11-05
Copied from_translation and from_scale functions to Matrix3x3

not-fl3:
Used this to construct transform matrices in 2D.

### DIFF
--- a/src/matrix.rs
+++ b/src/matrix.rs
@@ -140,6 +140,35 @@ impl<S: BaseFloat> Matrix3<S> {
         }
     }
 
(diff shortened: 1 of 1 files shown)


Rhuagh:
I'm not convinced this is a good idea, it might be confusing for users that use Matrix3 for its 3D rotational properties.

not-fl3:
Totally agree that Matrix3 impl is inappropriate and confusing place for those methods now. 

For my use-case I ended up with special local trait for 2d, inspired by glm's GLM_GTX_matrix_transform_2d header. Unfortunately no idea how to fit this approach well with cgmath's design :(

PR #475, opened 2019-01-12
Declare vector constructors to be const

nstoddard:
This makes it easier to create vectors in constants.

### DIFF
--- a/src/euler.rs
+++ b/src/euler.rs
@@ -78,7 +78,7 @@ use num::BaseFloat;
 #[repr(C)]
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-pub struct Euler<A: Angle> {
+pub struct Euler<A> {
     /// The angle to apply around the _x_ axis. Also known at the _pitch_.
     pub x: A,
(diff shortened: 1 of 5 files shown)


kvark:
Could you please expand this to other types, so that our `const` usage is more consistent?

nstoddard:
Done. Unfortunately the only other types that `const` can be added to at the moment are points, because for the other types the compiler gives an error:

```
error: trait bounds other than `Sized` on const fn parameters are unstable                               
  --> src/matrix.rs:84:6                                                                                 
   |                                                                                                     
84 | impl<S: BaseFloat> Matrix2<S> {                                                                     
   |      ^                                                                                              
```

nstoddard:
Actually, it looks like this might be doable for other types by removing the trait bounds from the constructors. I'll look into that tomorrow.

nstoddard:
I updated this PR to declare the rest of the constructors to be const.

Since this change requires removing trait bounds, it has the side effect that it's now technically possible to create, for instance, an `Euler<T>` for any type, not just angles. This probably isn't a problem because such a type can't accidentally be used as if it actually contained angles (also, types like `Vector2` could already be used to hold any type in a similar manner).

kvark:
Thank you!
bors r+

PR #483, opened 2019-05-02
Make `rand` dependency optional (but enabled by default)

### LINKED ISSUE #481: Make `rand` implementations optional (with a Cargo feature)
A crate of mine that depends on `cgmath` currently has this as part of its dependency graph:

```
├── cgmath v0.17.0
│   ├── approx v0.3.2
│   │   └── num-traits v0.2.6
│   ├── num-traits v0.2.6 (*)
│   └── rand v0.6.5
│       ├── libc v0.2.51
│       ├── rand_chacha v0.1.1
│       │   └── rand_core v0.3.1
│       │       └── rand_core v0.4.0
│       │   [build-dependencies]
│       │   └── autocfg v0.1.2
│       ├── rand_core v0.4.0 (*)
│       ├── rand_hc v0.1.0
│       │   └── rand_core v0.3.1 (*)
│       ├── rand_isaac v0.1.1
│       │   └── rand_core v0.3.1 (*)
│       ├── rand_jitter v0.1.3
│       │   └── rand_core v0.4.0 (*)
│       ├── rand_os v0.1.3
│       │   ├── libc v0.2.51 (*)
│       │   └── rand_core v0.4.0 (*)
│       ├── rand_pcg v0.1.2
│       │   └── rand_core v0.4.0 (*)
│       │   [build-dependencies]
│       │   └── autocfg v0.1.2 (*)
│       └── rand_xorshift v0.1.1
│           └── rand_core v0.3.1 (*)
│       [build-dependencies]
│       └── autocfg v0.1.2 (*)
```

No other crate depends on `rand`, so `cgmath` is pulling in roughly 9 crates into my dependency graph. I do not use the implementation of `rand` crates of `cgmath`, so this is wasted compilation time for me. 

Could those implementation put behind a `rand` feature? It can be enabled by default so that this change can be released as a minor version.

LukasKalbertodt:
Closes #481 

This PR leaves the feature enabled by default so that this can be released as a minor version soon. (I would suggest to disable it by default in future versions, but that's another discussion.)

The changes are pretty straight forward with one catch: I changed some `use` statements to nested imports (otherwise, there would be even more `#[cfg(feature = "rand")]` lines). Nested imports were stabilized in 1.25 (see [the edition guide on this feature](https://doc.rust-lang.org/edition-guide/rust-2018/module-system/nested-imports-with-use.html)). I'm not sure how cgmaths policy on minimum compiler version is. Maybe cgmath already requires a >= 1.25 compiler for other reasons. If my change is a problem, just tell me and I will change the imports back.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,9 +21,10 @@ matrix:
     - rust: nightly
 
 script:
-  - cargo build --no-default-features --features "$CARGO_FEATURES"
-  - cargo test --no-default-features --features "$CARGO_FEATURES"
+  - cargo build --features "$CARGO_FEATURES"
+  - cargo test --features "$CARGO_FEATURES"
   - |
     if [[ "$TRAVIS_RUST_VERSION" == "nightly" ]]; then
-      cargo bench --features "$CARGO_FEATURES"
+      # The benchmark always needs the 'rand' feature
+      cargo bench --features "$CARGO_FEATURES rand"
     fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ cache: cargo
 
 env:
   - CARGO_FEATURES=""
-  - CARGO_FEATURES="mint serde"
+  - CARGO_FEATURES="mint serde rand"
 
 matrix:
   include:
@@ -21,8 +21,8 @@ matrix:
     - rust: nightly
 
 script:
-  - cargo build --features "$CARGO_FEATURES"
-  - cargo test --features "$CARGO_FEATURES"
+  - cargo build --no-default-features --features "$CARGO_FEATURES"
+  - cargo test --no-default-features --features "$CARGO_FEATURES"
   - |
     if [[ "$TRAVIS_RUST_VERSION" == "nightly" ]]; then
       cargo bench --features "$CARGO_FEATURES"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,6 @@ keywords = ["gamedev", "math", "matrix", "vector", "quaternion"]
 name = "cgmath"
 
 [features]
-default = ["rand"]
 unstable = []
 swizzle = []
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ keywords = ["gamedev", "math", "matrix", "vector", "quaternion"]
 name = "cgmath"
 
 [features]
+default = ["rand"]
 unstable = []
 swizzle = []
 
@@ -24,7 +25,7 @@ swizzle = []
 approx = "0.3"
 mint = { version = "0.5", optional = true }
 num-traits = "0.2"
-rand = "0.6"
+rand = { version = "0.6", optional = true }
 serde = { version = "1.0", features = ["serde_derive"], optional = true }
 simd = { version = "0.2", optional = true }
 
--- a/src/angle.rs
+++ b/src/angle.rs
@@ -20,9 +20,11 @@ use std::f64;
 use std::iter;
 use std::ops::*;
 
-use rand::Rng;
-use rand::distributions::{Distribution, Standard};
-use rand::distributions::uniform::SampleUniform;
+#[cfg(feature = "rand")]
+use rand::{
+    Rng,
+    distributions::{Distribution, Standard, uniform::SampleUniform},
+};
 use num_traits::{cast, Bounded};
 
 use structure::*;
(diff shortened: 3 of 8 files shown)


kvark:
Looks fine by me overall.
Is non-rand case being tested by CI?

Leaving for others to have a look as well.

LukasKalbertodt:
> Is non-rand case being tested by CI?

No, currently not. 

We probably would like to have a test case without any optional features (without `rand`, `serde` and `mint`) and one with all enabled?  Testing all permutations (times two because of beta/stable) is probably overkill. Shall I change the CI configuration as described?

kvark:
Just doing `cargo test --no-default-features` would be fine by me.

LukasKalbertodt:
@kvark I just pushed a commit. I hope this is what you had in mind.

.travis.yml
```diff
@@ -21,8 +21,8 @@ matrix:
     - rust: nightly
 
 script:
-  - cargo build --features "$CARGO_FEATURES"
-  - cargo test --features "$CARGO_FEATURES"
+  - cargo build --no-default-features --features "$CARGO_FEATURES"
+  - cargo test --no-default-features --features "$CARGO_FEATURES"
```
kvark: this would *only* test with selected features. We need a test without any features, at least

LukasKalbertodt: I'm not quite sure what you mean. 

- In one row in the Travis "matrix" is with `CARGO_FEATURES=""`. And `cargo build --no-default-features --features ""` builds without any features, right? 
- In the other row of the matrix we test with `CARGO_FEATURES="mint serde rand"`, resulting in `cargo build --no-default-features --features "mint serde rand"`.

So as far as I understand it, it is tested once with `rand` and once without.

kvark: ah right, I missed that, sorry about the confusion!

LukasKalbertodt:
> anyone who depends on `cgmath` with no default features...

As `cgmath` didn't offer any default features before, I doubt many people use `cgmath` in that configuration. Furthermore RFC 1105 (which defines which changes are considered "major changes"), states that ["Altering the use of Cargo features" is a minor change](https://rust-lang.github.io/rfcs/1105-api-evolution.html#minor-change-altering-the-use-of-cargo-features).

kvark:
@LukasKalbertodt thank you for the link! Upon reading it carefully, I don't think it applies to our case. It's more about what features of *dependent* projects our project decides to enable. For example, we depend on `serde` with `serde_derive` feature enabled. If we were to change that list of required `serde` features in any way (what is referred to as "altering"), that wouldn't be a breaking change. However, changing the default features of self *is* a breaking change.

LukasKalbertodt:
@kvark Might be true :thinking: 

I guess you don't plan on releasing a major new version of `cgmath` soon? If not, I would probably need to use a fork until it is released. That wouldn't be a huge problem, but still unfortunate. 

And regarding this PR: can this be merged regardless of its breaking-nature? Or should I create a new PR against a different branch?

kvark:
@ozkriff @tomaka @brendanzab @csherratt @torkleyy 

I was about to say that I'm not intimately familiar with cgmath workflow and release procedures, and that we should ask somebody else, but then I scanned the history of merged PRs and mostly saw myself lately :( yikes! I hope I'm not the last holdout of `cgmath`, since my focus is graphics anyway, and I just want the ecosystem to be healty.

Anyhow, in other projects I have, the `master` is always one receiving breaking changes like this, while post-release patches go into release branches. I don't see any such branches here, but at least I see the release tags. So do I understand correctly that this project prefers a branch spawned from a release tag, merged, published, and a new tag produced, while the branch gets removed?

@LukasKalbertodt typically, releasing a breaking version upon seeing the first breaking change would waste a lot of effort by the community in updating... (unless they rely on [mint](https://github.com/kvark/mint)). So what I do is - scanning the list of issues and looking for other breaking changes that we accepted and haven't got merged/implemented yet. This is work, and I don't expect you to follow-up on it, but that's the process that I find working best.

For this PR, since we are breaking the API anyway, let's not enable `rand` by default.

LukasKalbertodt:
@kvark Sounds reasonable. I pushed a commit removing `rand` from default features. 

In general, I could see myself helping out to work on and co-maintain `cgmath`, in case there aren't enough people available. However, I have to hand in my master thesis (that project is using `cgmath` btw) in one month, so I don' exactly have a lot of time to spare. But afterwards, I'm available again, if there is interest from your side.

PR #488, opened 2019-08-12
bump rand -> 0.7, fix unused macro warnings in tests

spearman:
Would it be possible to get a new release published? Rand is still a default feature in the published version.


kvark:
This can't be published as a patch release, since the traits implemented by this code are different (rand 0.7 versus the previous)...

kvark:
That is to say, the code is good. Thank you for the PR!

PR #492, opened 2019-09-01
Clean up specialization and simd

NNemec:
The current SIMD implementation is based on the deprecated "simd" package and requires the unstable feature "specialization". So far, this was handled by significant code duplication and added complexity, making everything harder to understand even when "simd" was not in use at all.

In order to migrate from "simd" to "packed_simd" it appears valuable to first clean up the existing "simd" code, see what is still valuable and then see how to migrate.

This change reduces duplication, introduces a clear distinction of the "simd" feature and the "specialization" feature and moves most "simd" related code into separate source files.

The code is pure refactoring, keeping all functionality unchanged.

Testing the "simd" feature requires the 1.32 toolchain and commenting out the glium dev dependency. With that all tests run successfully.


kvark:
@NNemec could you add a good fat entry to the CHANGELOG describing the essence of this change as well?
@brendanzab know anybody who'd be willing to review this in more detail?

NNemec:
I did already add a CI test using 2019-01-01-nightly with feature "simd" which is passing.
I will write a changelog and perhaps a paragraph about the current state of SIMD somewhere.
For reviewing this change, I would suggest looking at the commits individually. I took great care committing the changes step by step.

kvark:
Yeah, I can see that your work is structured very well 👏 
Just not used to review SIMD code, so it will take me some time.

kvark:
Thanks!
bors r+

PR #500, opened 2020-04-14
Quaternion memory layout changed to `[x, y, z, w]`

aloucks:
The `From` and `Into` impls for `[S; 4]` and `(S, S, S, S)` have been changed accordingly. This is consistent with other libraries like glm, nalgebra, and glam, as well as the glTF spec.

Note that the `Quaternion::new` constructor has not yet been updated.

Related: https://github.com/rustgd/cgmath/issues/499

@kvark I hadn't previously thought much about the From/Into conversions for `[S; 4]` and `(S, S, S, S)`. It makes me a little nervous since this change is going to break safe code. We should probably also update the `Quaternion::new` constructor but I wanted to see if there was any feedback first.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
    default_fn! macro to reduce code duplication and complexity. Currently
    only needed for non-functional SIMD feature.
  - Refactored SIMD code into separate source files. See README.md for details.
+ - **Breaking**: Quaternion memory layout changed to `[x, y, z, w]`. The
+   `From` and `Into` impls for `[S; 4]` and `(S, S, S, S)` have been changed
+   accordingly.
+
 
 ### Added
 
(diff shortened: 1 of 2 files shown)


kvark:
I think that relying on [S; 4] be *any* representation is smelly. It's intended to just be a serialization form, only compatible with cgmath itself. If the user needs a particular form, they need to build it, or rely on `mint`, which is the intermediate.

With this position, it doesn't feel to me that cgmath needs any changes on how slices are converted. Nobody should rely on that implicitly anyway, and if they do - we are going to break their code.

aloucks:
> With this position, it doesn't feel to me that cgmath needs any changes on how slices are converted. Nobody should rely on that implicitly anyway, and if they do - we are going to break their code.

The slice layout needs to match the struct layout. There's a bunch of unit tests that enforce this and are broken if the slice and tuple conversions aren't updated. I would find it conceptually confusing if the slice/array didn't match the in memory representation to be honest.

kvark:
That's a good point, I didn't realize it's repr(C).

PR #504, opened 2020-06-03
Remove Float bound from MetricSpace and InnerSpace; move methods requ…

### LINKED ISSUE #470: dot product missing for Vector3<isize> ?
porting simple function from nalgebra:

```rust
let scale: isize = 2 * normal.0.dot(&self.0);
```

fails with:

```
error[E0599]: no method named `dot` found for type `cgmath::vector::Vector3<isize>` in the current scope
  --> src/vel.rs:26:41
   |
26 |         let scale: isize = 2 * normal.0.dot(&self.0);
   |                                         ^^^
   |
   = note: the method `dot` exists but the following trait bounds were not satisfied:
           `cgmath::vector::Vector3<isize> : cgmath::structure::InnerSpace`
```

Is there a reason why I can't take the dot product of isize vectors? (which is defined and mathematically sensible)

Or am I missing something and need to import something?

nstoddard:
…iring Float to new traits

This moves methods such as `distance` and `magnitude` to new traits called `MetricSpaceFloat` and `InnerSpaceFloat`.

This makes it possible to call methods like `dot` on vectors containing integers.

This fixes #470.

### DIFF
--- a/src/structure.rs
+++ b/src/structure.rs
@@ -195,7 +195,7 @@ where
 /// Examples are vectors, points, and quaternions.
 pub trait MetricSpace: Sized {
     /// The metric to be returned by the `distance` function.
-    type Metric: Float;
+    type Metric;
 
     /// Returns the squared distance.
     ///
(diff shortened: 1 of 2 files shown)


nstoddard:
Actually, please hold off on reviewing this; it should be possible to do this without adding new traits.

nstoddard:
Updated this PR to avoid adding new traits.

kvark:
Note that there is some bad happening on nightly CI: https://travis-ci.org/github/rustgd/cgmath/jobs/694107470#L549
We need to either fix CI signaling, or move to Github actions...
and we need to fix that Rand usage

Issue #490, opened 2019-08-24
Migrate to packed_simd


NNemec:
I found that cgmath has optional support for simd, which does not compile since Rust 1.33 and is deprecated in favor of packed_simd (see https://crates.io/crates/simd)

From what I can see, packed_simd is significantly different from simd, so migrating may be a tremendous effort. Has anyone looked into this? Is it feasible at all?

Boscop:
Any update on this? :)
Or is there another crate that supports simd that is optimized for 4x4 matrices?

NNemec:
I did spend some more time on this last year, but got stuck on a fundamental problem and ultimately moved on to other projects:
 
The problem is that fully optimized SIMD code requires proper memory alignment of the data. For a vector4, this can be easily achieved with packed_simd. For the far more common vector3, one field has to be added for padding, affecting the data layout of cgmath in general. To make matters worse, the content of that additional field has to be considered carefully, because even if you never look at it, some operations are affected by it in their performance (e.g. when dividing by zero, handling denormals INF or NAN).
 
One approach would be to guarantee that the all safe operations make sure that the padding field is always zero and assume for their output and assume that it is zero on their input. For most operations, that is trivial. Some operations require a bit of extra code, but that is worth it for simplicity.
 
Another approach would be to make the extra field explicit in the API and give the user control and responsibility over what is stored in it.
 
Further problems are the handling of vector2 and vector1 - each of which has to be considered individually for SIMD, even though it is questionable how much effort this is worth at all.
 
So there are some questions to be answered and lots of work to do, but I still think it is a realistic and valuable endeavor. However, I am personally focused on other topics at the moment and have no idea when I might get back to this.
 
Last I looked, I could not find any Rust library offering comparable functionality to cgmath with reasonable SIMD support, but I would be quite interested to hear whether I missed anything or the situation has changed.

aloucks:
There are a couple other math libraries with SIMD support:

https://github.com/termhn/ultraviolet
Ultraviolet's claim-to-fame is using "Structure of Arrays" (SoA) to gain a performance edge, but it may complicate your architecture (note the SoA types aren't mandatory). Ultraviolet seems quite experimental and lightly maintained. It uses rotors instead quaternions for rotation. 

https://github.com/bitshifter/glam-rs
Glam seems to be a bit more popular and I believe it to be a solid choice. I think it's inspired by DirectXMath, but with a smaller API surface area.

https://github.com/aloucks/directx_math
[...]

PR #505, opened 2020-06-03
Fix compilation of benchmarks

nstoddard:
This adds the `rand_isaac` dependency since that's no longer in the `rand` crate, but in the future we should consider switching to an RNG in `rand` so we don't need another dependency.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,8 @@ swizzle = []
 approx = "0.3"
 mint = { version = "0.5", optional = true }
 num-traits = "0.2"
-rand = { version = "0.7", optional = true }
+# small_rng used only for benchmarks
+rand = { version = "0.7", features = ["small_rng"], optional = true }
 serde = { version = "1.0", features = ["serde_derive"], optional = true }
 simd = { version = "0.2", optional = true }  # works only in rust toolchain up to 1.32
 
--- a/benches/common/macros.rs
+++ b/benches/common/macros.rs
@@ -19,7 +19,7 @@ macro_rules! bench_binop {
         fn $name(bh: &mut Bencher) {
             const LEN: usize = 1 << 13;
 
-            let mut rng = IsaacRng::from_entropy();
+            let mut rng = SmallRng::from_entropy();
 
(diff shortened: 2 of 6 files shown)


Cargo.toml
```diff
@@ -25,6 +25,7 @@ approx = "0.3"
 mint = { version = "0.5", optional = true }
 num-traits = "0.2"
 rand = { version = "0.7", optional = true }
+rand_isaac = { version = "0.2.0", optional = true } # Used only for benchmarks
```
kvark, line 28: should be in dev-dependencies then?

nstoddard: Done.

kvark:
Note: there is another error on nightly that's still happening:
https://travis-ci.org/github/rustgd/cgmath/jobs/694474980#L222

Could this PR address this one too? Looks related

nstoddard:
It looks like that error is due to the use of an old version of rust nightly (which is needed for SIMD support) which `rand_core` doesn't support. Optional dev-dependencies aren't supported, so one way to fix this would be to move `rand_isaac` back to a regular dependency so it can be enabled for benchmarks but not tests, but that's not ideal since it makes it possible for other crates to enable that dependency.

Another way would be to switch to an RNG that doesn't require another dependency. Are any in `rand` suitable for benchmarks?

kvark:
Did you consider using https://docs.rs/rand/0.7.3/rand/rngs/struct.SmallRng.html for this?

nstoddard:
`SmallRng` requires a feature to be enabled, which would then also be enabled for any crates that enable cgmath's `rand` feature; is that okay? If not, would `StdRng` work?

kvark:
I think we can enable it for rand, yes. Presumably, other crates enabling `rand` would be interested in using this.

nstoddard:
Done.

kvark:
Thank you! let's just see what CI thinks :)

kvark:
SIMD is still broken - https://travis-ci.org/github/rustgd/cgmath/jobs/694761384#L220

nstoddard:
That's a different one than the one that was broken by this PR (current nightly rather than nightly-2019-01-01), and it won't work until #490 is implemented.

kvark:
Yeah, I figured, hence this PR is merged :)

Issue #502, opened 2020-05-16
Why does `Point` exist?


InnocentusLime:
I've been re-reading the docs for `cgmath` and I honestly don't understand why there are both `Vector` and `Point` in the library. 

In pure maths these two concepts are barely distinguished, while in `cgmath` they are strictly split by introducing two `struct`s. Not only that, these structs have different functionality implemented for them which consequently makes the crate more complicated. What was the motivation behind that partition?

P.S.
I personally feel like all the functionality for `Vector` and `Point` should be implemented for one struct --- `Vector` (and `Point` should be depreacted). That will drastically simplify the crate and can fix issue `#501`

elrnv:
I think traditionally having Point be distinct from Vector is an effort to encode the distinction between affine spaces and euclidean spaces into the type system. This is usually done to prevent certain types of bugs (for example adding two points together is invalid, but subtracting them is perfectly fine).

I think the graphics community is somewhat split on this topic because each choice has its tradeoffs.

FWIW I think that `nalgebra` already fills the niche of dense linear algebra with strict types, so probably `cgmath` can play the role of a simpler API as you propose.

InnocentusLime:
I see. Well, I guess I'll create a PR for that.

nstoddard:
When I switched to cgmath I expected the distinction between `Point` and `Vector` to prevent some bugs in my code, but in practice I don't think I've encountered many, if any, situations where it actually prevented a bug. On the other hand, I've found many cases where I've had to convert between points and vectors, so I'm somewhat in favor of merging the two (only somewhat because the distinction does make sense mathematically).

kvark:
Just wanted to explicitly chime in here to say that your feedback is much welcome, and I agree that the niche for rich typing is already taken, and we can have `cgmath` with simpler types.

LukasKalbertodt:
Just to throw my opinion out there, too: I very much appreciate the distinction between `Point` and `Vector` in `cgmath`. In fact, it's one of the biggest positive things about the library IMO. The following is not meant in a threatening way, but if cgmath removes `Point`, I will probably switch to another library which has two types. And for me, the distinction of the two types actually prevented some real world bugs.

Instead of thinking about removing `Point`, maybe just try to make it easier to convert between them. One example where the distinction "annoyed" me when I had to calculate the centroid of some points (I am aware `cgmath` has a function for that, but it only works for a slice of points, not iterators). But yeah, even the current API is fine in my book.

[...]

PR #506, opened 2020-06-06
first steps to removal of `Point`

InnocentusLime:
I did the first steps to remove the `Point` (solving issue #502)
As result all the `PointN` structs were removed (and the `point.rs` file was removed too) and replaced with `VectorN`. `VectorN` had `EucledeanSpace` implemented too. Some associated types and methods were removed for simplification 

1. `Diff`, `Scalar` in `EuclideanSpace` 
2. `dot`, `from_vec` and `to_vec` for `EucledeanSpace`
3. `to_homogenous` and `from_homogenous` were added to `Vector3` for convenienve

The library seems to compile in my case. Most likely the change is too radical and depreacating should be considered instead

I wasn't sure about some methods in `Transform`, so I left them intact


src/vector.rs
```diff
@@ -426,6 +433,14 @@ impl<S: BaseNum> Vector2<S> {
 }
 
 impl<S: BaseNum> Vector3<S> {
+    #[inline]
+    pub fn to_homogeneous(self) -> Vector4<S> {
```
nstoddard, line 437: Points and vectors have different representations in homogeneous coordinates (vectors have a `w` coordinate of 0 so they're not affected by translations, points have a `w` coordinate other than 0 (often 1)), so these methods should probably be called `point_to_homogeneous` etc or something similar, or made private.

src/vector.rs
```diff
@@ -426,6 +433,14 @@ impl<S: BaseNum> Vector2<S> {
...
+        self.extend(S::one())
+    }
+    #[inline]
+    pub fn from_homogeneous(v : Vector4<S>) -> Vector3<S> {
+        v.truncate()
```
nstoddard, line 442: This needs to divide by the `w` coordinate.

PR #509, opened 2020-06-23
Allow inverted near far for inverted depth fp32

giordi91:
While trying to make an inverted depth using a floating-point depth buffer. I got a panic from the asserts! macro preventing such behaviour. 

I have removed such macros to allow for a higher accuracy inverted depth fp32 buffer. 
Here below a render doc screenshot of it working:
image

Tests still pass accordingly. 

I did not find any particular guideline on the contribution or how the PR should be formatted, so I hope this is good enough.

### DIFF
--- a/src/projection.rs
+++ b/src/projection.rs
@@ -117,8 +117,9 @@ impl<S: BaseFloat> From<PerspectiveFov<S>> for Matrix4<S> {
             "The vertical field of view cannot be greater than a half turn, found: {:?}",
             persp.fovy
         );
+
         assert!(
-            persp.aspect.abs() > S::zero(),
+            abs_diff_ne!(persp.aspect.abs(), S::zero()),
             "The absolute aspect ratio cannot be zero, found: {:?}",
             persp.aspect.abs()
         );
@@ -133,7 +134,7 @@ impl<S: BaseFloat> From<PerspectiveFov<S>> for Matrix4<S> {
             persp.far
         );
         assert!(
-            (persp.far - persp.near).abs() > S::epsilon(),
+            abs_diff_ne!((persp.far - persp.near).abs() ,S::zero()),
             "The far plane and near plane are too close, found: far: {:?}, near: {:?}",
             persp.far,
             persp.near
--- a/src/projection.rs
+++ b/src/projection.rs
@@ -118,6 +118,11 @@ impl<S: BaseFloat> From<PerspectiveFov<S>> for Matrix4<S> {
             persp.fovy
         );
         assert!(
+            persp.aspect.abs() > S::zero(),
+            "The absolute aspect ratio cannot be zero, found: {:?}",
+            persp.aspect.abs()
+        );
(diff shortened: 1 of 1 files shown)


src/projection.rs
```diff
@@ -132,12 +127,6 @@ impl<S: BaseFloat> From<PerspectiveFov<S>> for Matrix4<S> {
             "The far plane distance cannot be below zero, found: {:?}",
             persp.far
         );
-        assert!(
-            persp.far > persp.near,
```
kvark, line 136: we should probably still assert if they are too close, or otherwise `        let c2r2 = (persp.far + persp.near) / (persp.near - persp.far);` would be screwed up

src/projection.rs
```diff
@@ -117,11 +117,6 @@ impl<S: BaseFloat> From<PerspectiveFov<S>> for Matrix4<S> {
             "The vertical field of view cannot be greater than a half turn, found: {:?}",
             persp.fovy
         );
-        assert!(
-            persp.aspect > S::zero(),
```
kvark, line 121: same here. It's fine to be negative, but it's not fine to be zero-ish

src/projection.rs
```diff
@@ -117,6 +117,11 @@ impl<S: BaseFloat> From<PerspectiveFov<S>> for Matrix4<S> {
             "The vertical field of view cannot be greater than a half turn, found: {:?}",
             persp.fovy
         );
+        assert!(
+            persp.aspect.abs() > S::zero(),
```
kvark, line 121: can we use the approx methods/macros here, like https://docs.rs/approx/0.3.2/approx/macro.assert_abs_diff_ne.html

src/projection.rs
```diff
@@ -127,6 +132,12 @@ impl<S: BaseFloat> From<PerspectiveFov<S>> for Matrix4<S> {
             "The far plane distance cannot be below zero, found: {:?}",
             persp.far
         );
+        assert!(
+            (persp.far - persp.near).abs() > S::epsilon(),
```
kvark, line 136: same here, let's use the approx methods more directly

giordi91:
Something like this? I did not use the assert_* variant because I did not seem to be able to pass in a message. I find reading the documentation of the macro being confusing for me since I am fairly new to Rust. Let me know if that fits the requirements

src/projection.rs
```diff
@@ -133,7 +134,7 @@ impl<S: BaseFloat> From<PerspectiveFov<S>> for Matrix4<S> {
             persp.far
         );
         assert!(
-            (persp.far - persp.near).abs() > S::epsilon(),
+            abs_diff_ne!((persp.far - persp.near).abs() ,S::zero()),
```
kvark, line 137: this should be `abs_diff_ne!(persp.far, persp.near)`

giordi91:
No worries, it is mostly me being clumsy with a new language and library. Thanks for your patience.

PR #511, opened 2020-07-19
Sized Angle trait

### LINKED ISSUE #510: Angle trait + Sized
`Angle` is defined as unsized trait, but all of its implementors are sized (i.e. `Rad` and `Deg`). Maybe it should be defines as sized then? It helps define structs with `Angle` on stack instead of passing by `&dyn Angle` or generic like `A: Angle<..> + Sized`.

r4v3n6101:
I don't really know how to deal with it, but seems it fixes #510.

### DIFF
--- a/src/structure.rs
+++ b/src/structure.rs
@@ -564,6 +564,7 @@ where
 ///
 pub trait Angle
 where
+    Self: Sized,
     Self: Copy + Clone,
     Self: PartialEq + cmp::PartialOrd,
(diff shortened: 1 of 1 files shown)


elrnv:
I believe `iter::Sum` inherits `Sized`, which makes `Angle` also `Sized`. So if IIUC this change has no effect on functionality, other than making it explicit that Angle is `Sized`. Is this right?

Issue #354, opened 2016-05-11
Remove `one` and `concat` methods from `Rotation` and `Transform`


bjz:
These should be inherited from `One`.

bjz:
This has the added benefit of allowing us to implement the rotation traits on matrices.

bjz:
Looking at it now, this will require a type signature of:

```rust
impl<P: EuclideanSpace, R: Rotation<P>> One for Decomposed<P::Diff, R>
```
This causes:

```
error: the type parameter `P` is not constrained by the impl trait, self type, or predicates [--explain E0207]
  --> src/transform.rs:60:6
60 |> impl<P: EuclideanSpace, R: Rotation<P>> One for Decomposed<P::Diff, R> where
   |>
```
We'll need to finally bite the bullet and convert `Rotation` and `Transform` to use associated types if we want to do this. So:

```
trait Rotation: Sized + One {
    type Vector: InnerSpace;
    type Point: EuclideanSpace<Diff = Vector>;

    ...
}

trait Transform: Sized + One {
    type Vector: InnerSpace;
    type Point: EuclideanSpace<Diff = Vector>;

    ...
}
```

PR #513, opened 2020-08-09
Now there is only one one

josh65536:
This fixes the `multiple one found` issue (one function mentioned in #354) with attempting to use `Matrix4::one()` or some other matrix's `one()` function when importing the prelude, which includes both `Transform` and `One`. The changes:
- `Transform::one()` is gone
- `Transform` is now a subtrait of `One`
- `Decomposed` implements `One`.
- `EuclideanRotation` is a new trait specifically for implementers of `Rotation<P>` for exactly one `P`. This allows `Decomposed` to implement `One` without running into unconstrained type parameter issues. All current implementers of `Rotation` in the `cgmath` crate implement `EuclideanRotation`.
- `Decomposed` implements `Mul` so it can implement `One`.

I know there was an attempt to do it in #386, but it was abandoned and closed.

There seems to be a few extra changes from running `cargo fmt`.

### DIFF
--- a/src/rotation.rs
+++ b/src/rotation.rs
@@ -61,6 +61,19 @@ where
     fn invert(&self) -> Self;
 }
 
+/// A trait for a rotation in a specific Euclidean space.
+/// Implement this trait for any type that implements `Rotation<P>`
+/// for exactly one `P`,
+/// setting `type Euclidean = P`.
+/// In particular, a rotation cannot be used in `Decomposed`
+/// unless it implements `EuclideanRotation`.
+pub trait EuclideanRotation: Rotation<<Self as EuclideanRotation>::Euclidean>
+where
+    <<Self as EuclideanRotation>::Euclidean as EuclideanSpace>::Scalar: BaseFloat,
+{
+    type Euclidean: EuclideanSpace;
+}
+
 /// A two-dimensional rotation.
 pub trait Rotation2<S: BaseFloat>:
     Rotation<Point2<S>> + Into<Matrix2<S>> + Into<Basis2<S>>
@@ -181,6 +194,10 @@ impl<'a, S: 'a + BaseFloat> iter::Product<&'a Basis2<S>> for Basis2<S> {
     fn product<I: Iterator<Item = &'a Basis2<S>>>(iter: I) -> Basis2<S> {
         iter.fold(Basis2::one(), Mul::mul)
     }
+}
+
+impl<S: BaseFloat> EuclideanRotation for Basis2<S> {
+    type Euclidean = Point2<S>;
 }
 
 impl<S: BaseFloat> Rotation<Point2<S>> for Basis2<S> {
@@ -332,6 +349,10 @@ impl<'a, S: 'a + BaseFloat> iter::Product<&'a Basis3<S>> for Basis3<S> {
     fn product<I: Iterator<Item = &'a Basis3<S>>>(iter: I) -> Basis3<S> {
         iter.fold(Basis3::one(), Mul::mul)
     }
+}
+
+impl<S: BaseFloat> EuclideanRotation for Basis3<S> {
+    type Euclidean = Point3<S>;
 }
 
 impl<S: BaseFloat> Rotation<Point3<S>> for Basis3<S> {
--- a/src/rotation.rs
+++ b/src/rotation.rs
@@ -42,14 +42,23 @@ where
     type Space: EuclideanSpace;
 
     /// Create a rotation to a given direction with an 'up' vector.
-    fn look_at(dir: <Self::Space as EuclideanSpace>::Diff, up: <Self::Space as EuclideanSpace>::Diff) -> Self;
+    fn look_at(
+        dir: <Self::Space as EuclideanSpace>::Diff,
+        up: <Self::Space as EuclideanSpace>::Diff,
+    ) -> Self;
 
     /// Create a shortest rotation to transform vector 'a' into 'b'.
     /// Both given vectors are assumed to have unit length.
-    fn between_vectors(a: <Self::Space as EuclideanSpace>::Diff, b: <Self::Space as EuclideanSpace>::Diff) -> Self;
+    fn between_vectors(
+        a: <Self::Space as EuclideanSpace>::Diff,
+        b: <Self::Space as EuclideanSpace>::Diff,
+    ) -> Self;
 
     /// Rotate a vector using this rotation.
-    fn rotate_vector(&self, vec: <Self::Space as EuclideanSpace>::Diff) -> <Self::Space as EuclideanSpace>::Diff;
+    fn rotate_vector(
+        &self,
+        vec: <Self::Space as EuclideanSpace>::Diff,
+    ) -> <Self::Space as EuclideanSpace>::Diff;
 
     /// Rotate a point using this rotation, by converting it to its
     /// representation as a vector.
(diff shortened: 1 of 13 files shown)


src/transform.rs
```diff
@@ -69,21 +67,50 @@ pub struct Decomposed<V: VectorSpace, R> {
     pub disp: V,
 }
 
-impl<P: EuclideanSpace, R: Rotation<P>> Transform<P> for Decomposed<P::Diff, R>
+// using `EuclideanRotation` here to avoid the possibility that
+// the same struct implements `Rotation<P1>` and `Rotation<P2>`
+// and the compiler has to guess what `P` is.
+impl<P: EuclideanSpace, R: EuclideanRotation<Euclidean = P>> One for Decomposed<P::Diff, R>
```
kvark, line 73: should this be `impl<R: EuclideanRotation> One for Decomposed<<R::Euclidean as Xxx>>::Diff, R>`?

kvark:
What is the problem (in more detail) that happens if you don't add the new trait?

src/transform.rs
josh65536: It could be that, but that would make all mentions of `P` a lot more verbose, and there are several of them.

josh65536:
If I wrote:
```rust
impl<P: EuclideanSpace, R: Rotation<P>> One for Decomposed<P::Diff, R>
where
    P::Scalar: BaseFloat,
    P::Diff: VectorSpace,
{
    fn one() -> Self {
        Decomposed {
            scale: P::Scalar::one(),
            rot: R::one(),
            disp: P::Diff::zero(),
        }
    }
}
```
then the compiler will complain that `P` is unconstrained.

This doesn't currently happen in `cgmath`, but imagine that there was some type `AnotherPoint2` that implemented `EuclideanSpace<Diff = Vector2<f64>>`. Also imagine that some type `RotationAny` implemented both `Rotation<Point2<f64>>` and `Rotation<AnotherPoint2<f64>>`. Then `Decomposed<Vector2<f64>, RotationAny>` implements `One` twice...

Actually, I just realized that in this case, since `R::one()` is coming from `One` and not `Rotation<P>`, so I can rewrite the body as
```rust
Decomposed {
    scale: P::Diff::Scalar::one(),
    rot: <R as One>::one(),
    disp: P::Diff::zero(),
}
```
and see that there is no dependence on `P`. I can *almost* replace `P: EuclideanSpace` with `V: VectorSpace` and replace `P::Diff` with `V`, but that won't deal with `Rotation` requiring `P` as a parameter. So here are some possibilities:
- The current pull request, with its upsides and downsides.
- Write it as `impl<V: VectorSpace, R: One> One for Decomposed<V, R>`. This eliminates the `EuclideanRotation` trait, but is also more general on `R` than I'd like.
- Move the associated type of `EuclideanRotation` into `Rotation` and remove the parameter on `Rotation`. This would make sense unless we want it to be possible for a type to implement `Rotation` for multiple Euclidean spaces. This eliminates the extra trait and allows me to write `impl<P: EuclideanSpace, R: Rotation<Euclidean=P>> One for Decomposed<P::Diff, R>`, constraining `P`. However, it could cause a lot of breakage.
- Write an RFC for relaxing the rules on type parameter constraints. In particular, it would be nice to write something like `impl<V: VectorSpace, R: exists<P: EuclideanSpace> Rotation<P>> One for Decomposed<V, R>` and simply not be able to make the implementation dependent on `P`. I don't know the chances of this getting approved.

kvark:
Thank you for elaborating here!
I like the (2) and (3) choices:

> Write it as impl<V: VectorSpace, R: One> One for Decomposed<V, R>. This eliminates the EuclideanRotation trait, but is also more general on R than I'd like.

What is the concern here, exactly? Is the generosity going to be a footgun in some situation?

> Move the associated type of EuclideanRotation into Rotation and remove the parameter on Rotation. This would make sense unless we want it to be possible for a type to implement Rotation for multiple Euclidean spaces.

cgmath needs to steer away from being generic into being more practical, so this constraint seems reasonable to me.

src/rotation.rs
```diff
@@ -30,53 +30,42 @@ use vector::{Vector2, Vector3};
...
 /// creates a circular motion, and preserves at least one point in the space.
-pub trait Rotation<P: EuclideanSpace>: Sized + Copy + One
+pub trait Rotation: Sized + Copy + One
 where
     // FIXME: Ugly type signatures - blocked by rust-lang/rust#24092
-    Self: approx::AbsDiffEq<Epsilon = P::Scalar>,
-    Self: approx::RelativeEq<Epsilon = P::Scalar>,
-    Self: approx::UlpsEq<Epsilon = P::Scalar>,
-    P::Scalar: BaseFloat,
+    Self: approx::AbsDiffEq<Epsilon = <<Self as Rotation>::Space as EuclideanSpace>::Scalar>,
```
kvark, line 36: I don't think you need `<Self as Rotation>::` anywhere. Can you just use `Self::`?

josh65536:
It turns out that the second option is blocked by a call to `rotate_vector` in the implementation of `Mul` for `Decomposed`. It would actually require `R` to implement `Rotation`. So option 3 is taken.
I noticed a few other traits like `Rotation2`, `Rotation3`, and `Transform` that have type parameters that should be associated types instead. It wouldn't make sense for the same type to implement both `Rotation2<f32>` and `Rotation2<f64>`, for example.

kvark:
> I noticed a few other traits like Rotation2, Rotation3, and Transform that have type parameters that should be associated types instead. It wouldn't make sense for the same type to implement both Rotation2<f32> and Rotation2<f64>, for example.

yes, agreed!

src/rotation.rs
josh65536: Nope, the compiler complains:
`error[E0391]: cycle detected when computing the bounds for type parameter 'Self'`

josh65536:
Should I save fixing those for another pull request or combine it with this one?

kvark:
I think it would make sense to combine them, if you can. Thank you!

src/rotation.rs
kvark: ugh

josh65536:
It turns out that `Matrix3<S>` implements both `Transform<Point2<S>>` and `Transform<Point3<S>>` since one of them allows translations and the other doesn't. However, `Rotation2`, `Rotation3`, `Transform2`, and `Transform3` didn't need their bounds.

kvark:
@aloucks @elrnv would you want to help reviewing these changes? They are pretty serious

src/transform.rs
```diff
@@ -69,21 +67,47 @@ pub struct Decomposed<V: VectorSpace, R> {
     pub disp: V,
 }
 
-impl<P: EuclideanSpace, R: Rotation<P>> Transform<P> for Decomposed<P::Diff, R>
+impl<P: EuclideanSpace, R: Rotation<Space = P>> One for Decomposed<P::Diff, R>
 where
     P::Scalar: BaseFloat,
-    // FIXME: Investigate why this is needed!
     P::Diff: VectorSpace,
```
elrnv, line 73: Can we remove this bound here?

src/rotation.rs
```diff
@@ -30,30 +30,41 @@ use vector::{Vector2, Vector3};
...
+    Self: approx::RelativeEq<Epsilon = <<Self as Rotation>::Space as EuclideanSpace>::Scalar>,
+    Self: approx::UlpsEq<Epsilon = <<Self as Rotation>::Space as EuclideanSpace>::Scalar>,
+    <<Self as Rotation>::Space as EuclideanSpace>::Scalar: BaseFloat,
```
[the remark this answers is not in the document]
josh65536: The first three are bounds on `Self`, but the last one can be simplified.

src/transform.rs
josh65536: It's needed because there's a lot of vector-scalar multiplication in the implementation.

elrnv: This is just a bound on `One` not on `Mul` right? It seems there are only vector-scalar multiplications in the `Mul` implementation.

josh65536: Oops, I was looking at the wrong line.

PR #514, opened 2020-08-11
Add Transform::{look_at_rh, look_at_lh} and deprecate look_at (continued)

elrnv:
This is a continuation of #508 with the following additions:
- `_lh` and `_rh` suffixes are explained in each affected function.
- `look_at` is renamed to `look_to` on Matrix2 and 2D rotations for consistency.
- Other uses of `look_at` were changed to `look_to` as necessary in the `Rotation` trait.

These changes make `look_at` and `look_to` functions consistent in meaning throughout the crate (with exception of deprecated functions).


src/matrix.rs
```diff
@@ -101,14 +101,29 @@ impl<S> Matrix2<S> {
...
+    /// `dir`. This is similar to `look_to`, but does not take an `up` vector.
     /// This will not cause `unit_y()` to flip when `dir` crosses over the `up` vector.
+    #[deprecated = "Use Matrix2::look_to_stable"]
     pub fn look_at_stable(dir: Vector2<S>, flip: bool) -> Matrix2<S> {
+        Matrix2::look_to_stable(dir, flip)
+    }
+
+    /// Create a transformation matrix that will cause `unit_x()` to point at
+    /// `dir`. `unit_y()` will be perpendicular to `dir`, and the closest to `up`.
+    pub fn look_to(dir: Vector2<S>, up: Vector2<S>) -> Matrix2<S> {
```
aloucks, line 119: This should be `look_to_{lh|rh}`. I didn't update these because I didn't know what the current implementation was.

src/quaternion.rs
```diff
@@ -474,7 +474,7 @@ impl<S: BaseFloat> From<Quaternion<S>> for Basis3<S> {
 
 impl<S: BaseFloat> Rotation<Point3<S>> for Quaternion<S> {
     #[inline]
-    fn look_at(dir: Vector3<S>, up: Vector3<S>) -> Quaternion<S> {
```
aloucks, line 477: There should be `{lh|rh}` variants.

src/matrix.rs
elrnv: Is there a handedness distinction for 2D rotations?

src/quaternion.rs
elrnv: I was hoping to consolidate the look functions in `Transform` and `Rotation` in a separate PR since it is more involved. This looks like it is at least has consistent naming. What do you think?

src/matrix.rs
aloucks: I think that there is because handedness defines the direction of rotation. However, I had the same (unresolved) question which is why I decided not to modify it in the other PR. My reasoning was that it's probably better to leave it alone until I (or someone else) has time to figure out the correct path forward.

src/quaternion.rs
aloucks: I guess my thinking is that it needs the handedness for it to be consistent and I don't think the implementations are consistent right now (i.e. they're not all LH or all RH and some of the implementations are on the raw structs instead of the traits).  

What I think needs to happen is that someone needs to go through the remaining `look_` function and determine if it's currently `lh` or `rh` and deprecate it in favor of an explicitly named version of that function. While checking these, the corresponding missing `lh` or `rh` version can also be added. I think you've done this for `Matrix3`, but the other ones are missing the handedness suffix. I skipped over the ones I wasn't sure of rather than just renaming them to `look_to` (without the suffix) because they'll just need to be renamed again. 

I was also thinking that it doesn't really make sense to have `look_at_` variants for structures that can only represent rotation. I think these should **only** have `look_to_` functions. We should probably sort out these kinds of questions before considering the removal of `Rotation`.

src/matrix.rs
elrnv: I'm confused. I thought handedness determines between coordinate systems. The direction of rotation should be indistinguishable here. I.e. the rotation is the same regardless of which way you turn. For 2D rotations, setting an upside-down up vector should cause the coordinate system to flip, which, I think does the equivalent of what the 3D handedness does. I.e. it flips the sign of the output matrix determinant. Maybe I'm missing something here? It seems like we can either do handedness here without the `up` vector or leave it as is.

src/quaternion.rs
elrnv: > We should probably sort out these kinds of questions before considering the removal of Rotation.
Agreed, so shall we punt that to another PR?
For this PR I am happy with having the look_at vs look_to consistency throughout cgmath.

kvark:
Could somebody sum up the status of this PR? Is there a consensus, and what's left to do?

elrnv:
IIRC this PR aims to get consistent naming for `_to` and `_at` variants of the `look_` functions on top of #508.

It does this to a certain extent without committing to major changes to the `Rotation` and `Transform` APIs. In particular, this PR changes `look_at` to `look_to` in `Rotation` without the `_lh` or `_rh` suffixes for now in hopes of properly resolving the overlap between the `Rotation` and `Transform` APIs in another PR since I don't have the time at the moment to dwell on this. **The remaining decision I think is whether to roll back the changes to the `Rotation` trait or keep it as is since it has better consistency for the time being,** and it's not clear that the next PR will come any time soon.

**The other point to resolve is whether Matrix2 needs `_lh` and `_rh` variants for the `look_to` function.** In my opinion it does not, but I am more than happy to either revert that change or do something else there.

#508 leaves the `look_to` and `look_at` function variants in an inconsistent state to (I think) avoid breaking the APIs of things like Rotation and Matrix2 temporarily before a more appropriate solution is found. Given the cadence of this crate, I think the changes in this PR are appropriate to get some consistency albeit not necessarily with respect to `_lh` and `_rh` variants of the `look_` functions.

thatcomputerguy0101:
I believe the current implementation of `look_at/to` in Matrix2 is suited for 2d spritework, where in order for a sprite to maintain an "upwards" orientation while looking in the opposite direction, it must mirror itself. It is a related, but different operation than in the Matrix3 and Matrix4 implementations. Even with that in mind, I would still favor renaming it to `look_to` instead of `look_at`.

PR #515, opened 2020-08-11
Fixed opposite quaternion slerp bug

### LINKED ISSUE #498: Quaternion::slerp gives different result than gl-matrix
I'm implementing skeletal animation and there were glitches for a very short time. After some debugging, it seems that a call to `slerp` gives different results:

```rs
use cgmath::Quaternion;

fn main() {
    let a = Quaternion::<f32>::new(0.00052311074, 0.9999999, 0.00014682197, -0.000016342687);
    let b = Quaternion::<f32>::new(0.019973433, -0.99980056, -0.00015678025, 0.000013882192);

    println!("{:?}", a.slerp(b, 0.5));
}
```

This prints:

```
Quaternion { s: 1.0001583, v: Vector3 [0.009726516, -0.0004859287, -0.000120063356] }
```

I've tried the same quaternions with gl-matrix:

```
const a = quat.fromValues(0.9999999, 0.00014682197, -0.000016342687, 0.00052311074);
const b = quat.fromValues(-0.99980056, -0.00015678025, 0.000013882192, 0.019973433);

const r = quat.create();
quat.slerp(r, a, b, 0.5);
console.log(r);
```

And this prints (the scalar part is the last component here):

```
Float32Array(4) [0.9999527335166931, 0.00015180908667389303, -0.000015113232620933559, -0.009725671261548996]
```

josh65536:
Fixes #498.
And oops, the same rustfmt collateral


src/quaternion.rs
```diff
@@ -126,30 +126,31 @@ impl<S: BaseFloat> Quaternion<S> {
...
-    pub fn slerp(self, other: Quaternion<S>, amount: S) -> Quaternion<S> {
-        let dot = self.dot(other);
-        let dot_threshold = cast(0.9995f64).unwrap();
+    pub fn slerp(self, mut other: Quaternion<S>, mut amount: S) -> Quaternion<S> {
+        let mut dot = self.dot(other);
+        let dot_threshold: S = cast(0.9995f64).unwrap();
+
+        // Fix for opposite quaternions: pick a perpendicular target and change the amount accordingly
+        // Therefore, the following code cannot assume that 0 <= amount <= 1.
+        if ulps_eq!(dot, -S::one()) {
```
kvark, line 135: why is -1 compared by `ulps_eq` but 1 is compared by the `dot_threshold` thing? Can we align these?

josh65536: 1 is compared by `dot_threshold` because in that range, `slerp` is close enough to `nlerp` to not matter. But in the case of -1, this is not the case, and so we should have a tighter threshold. If the vectors are very close to being on a line but the dot product isn't quite -1, their span is still a plane and `slerp` can still give a good result.

kvark: Hmm. A quaternion is the same if you multiply all the components by -1. So if the dot becomes negative, you may as well do that operation on one of the operands, and you get a positive dot product. Can we unify therefore the check for proximity to 1 and -1 ?

josh65536: No, because a negative dot product represents a rotation of more than 180 degrees, and `slerp` should respect that.

kvark: Ok, I'm genuinely confused now. When interpolating between any 2 directions, and choosing the shortest arc, there shouldn't be a rotation exceeding 180 degrees.

josh65536:
Not sure why the check failed. `q` and `r` are `Quaternion<f64>`, so `q.slerp(r, 0.25)` is a `Quaternion<f64>`, and thus `q.slerp(r, 0.25).dot(q)` is a `<Quaternion<f64> as VectorSpace>::Scalar`, which is `f64`. `0.5f64.sqrt()` is quite obviously a `f64`. So there is no `f32` to be found.

src/quaternion.rs
josh65536: That's not what `slerp` does, though. It takes the shortest arc only if the dot product is nonnegative. There could be a `shortest_slerp` function that always takes the shortest arc.

kvark: Interesting. So looking at the example code on the [wiki](https://en.wikipedia.org/wiki/Slerp), they appear to be doing precisely what I'm asking for:
```cpp
    // If the dot product is negative, slerp won't take
    // the shorter path. Note that v1 and -v1 are equivalent when
    // the negation is applied to all four components. Fix by 
    // reversing one quaternion.
    if (dot < 0.0f) {
        v1 = -v1;
        dot = -dot;
    }
```

josh65536: Wouldn't that need to be done for `nlerp` as well for consistency? And I notice that in [this post](http://number-none.com/product/Understanding%20Slerp,%20Then%20Not%20Using%20It/), the author doesn't do that reversing.

kvark: good question! it sounds like we should do that for nlerp as well

josh65536: I made them both take the shortest path now.

aloucks:
All you need to do is check if the dot product is negative. If it is negative, negate the quaternion and recompute the dot product again.

PR #516, opened 2020-08-30
Optional JSON schema generation

afnanenayet:
Allow users to optionally derive JSON schemas using structs from this crate if they include `schemars`. This will allow consumers of this library to generate JSON schemas from structs that include structs created in this crate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ num-traits = "0.2"
 rand = { version = "0.7", features = ["small_rng"], optional = true }
 serde = { version = "1.0", features = ["serde_derive"], optional = true }
 simd = { version = "0.2", optional = true }  # works only in rust toolchain up to 1.32
+schemars = { version = "0.7", optional = true }
 
 [dev-dependencies]
 # glium = "0.23"  # causes problems with nightly-2019-01-01 needed for testing "simd"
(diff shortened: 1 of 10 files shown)


kvark:
Thank you for making this PR!

I honestly haven't worked with JSON schemas in Rust. And I don't know what the established conventions are, so it's hard to tell if `schemars` would be needed in cgmath.
Given that cgmath is fairly low-supported atm, I'm curious to see if other libraries (with better support and stuffing) investigated their part. Did nalgebra, glam, and whatever, came to any conclusion wrt `schemars` integration?

afnanenayet:
It doesn't seem like either of the libraries mentioned have dealt with schema generation yet, at least not that I was able to find.

My use case is that I'm serializing structs that have `Vector3` as a field, and I want to derive the JSON schema for those structs, and I can't do that unless every underlying struct also derives the schema.

kvark:
What are the stability guarantees for `schemars`? Would it work if the struct has fields serialized by different versions of this crate?

afnanenayet:
I'm actually not entirely sure, the documentation says this:

> One of the main aims of this library is compatibility with Serde. Any generated schema should match how serde_json would serialize/deserialize to/from JSON.

`serde_json` and `serde` seem incredibly stable, so if this is piggybacking off of that, then I think this shouldn't have issues. I'll try to dig around the `schemars` source and see what's going on.

PR #519, opened 2020-12-05
Add github workflows

aloucks:


.github/workflows/tests.yml
```diff
@@ -0,0 +1,87 @@
...
+  #         command: fmt
+  #         args: --all -- --check
+  test:
+    name: Test
+    runs-on: ${{ matrix.os }}
+    env:
+      RUST_BACKTRACE: 1
+    strategy:
+      matrix:
+        os: [windows-latest, ubuntu-latest, macos-latest]
```
kvark, line 36: that's a whole lot of build: 3 * 2 * 2 in total. Can we perhaps mix and match those into fewer builds?

kvark (commented):
Thank you for the PR!
There is one concern about the GHA setup. Plus, please remove the travis yaml

.github/workflows/tests.yml
aloucks: I was actually considering adding _more_ builds by adding all of the different permutations of features (i.e. none, mint, serde, mint+serde). This is what the strategy matrix is for :)

I could certainly remove the windows and macos builds, but is it actually an improvement? Yes there are `3 * 2 * 2` total builds here, but they run in parallel and the entire CI process finishes in under 3 minutes. I don't consider this to be excessive. I'd much prefer to run CI on all the major platforms and only reduce this if there's an issue with long build queues.

PR #520, opened 2020-12-05
Release 0.18

kvark:
Closes #499
r? @aloucks  (check the invites on github and crates)

PR #521, opened 2020-12-05
Update readme build status and remove IDE config

aloucks:


kvark:
This partially collides with #520 
also, we are now getting the spam of warnings:
> use of deprecated associated function `matrix::Matrix4::<S>::look_at_dir`: Use Matrix4::look_to_rh

PR #523, opened 2020-12-24
Upgrade approx and rand crates

fintelia:


Cargo.toml
```diff
@@ -21,11 +21,11 @@ unstable = []
 swizzle = []
 
 [dependencies]
-approx = "0.3"
+approx = "0.4"
 mint = { version = "0.5", optional = true }
 num-traits = "0.2"
 # small_rng used only for benchmarks
-rand = { version = "0.7", features = ["small_rng"], optional = true }
+rand = { version = "0.8", features = ["small_rng"], optional = true }
```
kvark, line 28: oh smack, now we'll need another breaking version release?

kvark (approved):

Cargo.toml
aloucks: What do you mean by _another_? `0.18` hasn't been released yet according to crates.io.

kvark: Sorry, 0.18 was in production for so long I forgot it still hasn't released. So this change is very much on time!

PR #525, opened 2021-02-12
rkyv Integration

MarekWojt:
This basically adds a feature for support for [djkoloski/rkyv](https://github.com/djkoloski/rkyv), a zero-copy deserialization framework for Rust.


kvark (approved):
Makes sense, thank you!

kvark:
We need to make sure that `rkyv` feature is tested on CI before merging.

PR #527, opened 2021-04-05
Simplify "fold_array" macro

Besler:
- Macro now applies to arbitrary dimensions
- Removes the reference format of the macro `(&$method:ident, { $x:expr })`. Tests passed fine, I couldn't find anywhere it was needed.

Macro expand to an inverted composition:

**Before**
```rust
#[inline]
fn sum(self) -> S
where
    S: Add<Output = S>,
{
    self.x.add(self.y).add(self.z)
}
```

**Now**
```rust
#[inline]
fn sum(self) -> S
where
    S: Add<Output = S>,
{
    self.x.add(self.y.add(self.z))
}
```

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -139,30 +139,8 @@ macro_rules! impl_assignment_operator {
 }
 
 macro_rules! fold_array {
-    (&$method:ident, { $x:expr }) => {
-        *$x
-    };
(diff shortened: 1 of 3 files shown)


kvark (approved):
Thank you!
bors r+

PR #528, opened 2021-04-20
Implement From instead of Into for array/tuple/mint conversions

agausmann:
Fixes missing impls of `impl From<ArrayN> for ForeignType` which is currently provided as `impl Into<ForeignType> for ArrayN`

From the [`std::convert::Into` documentation](https://doc.rust-lang.org/stable/std/convert/trait.Into.html):

> One should avoid implementing Into and implement From instead. Implementing From automatically provides one with an implementation of Into thanks to the blanket implementation in the standard library.

While it isn't very significant, as you can write `cgmath_vector.into()` in most places, it makes code cleaner where type inference is not possible (e.g. `format!` macros):

```rust
// With `Into`: 
<Vector2<f32> as Into<[f32; 2]>>::into(a_vector)

// ... versus with `From`:
<[f32; 2]>::from(a_vector)
```

  (Example adapted from this [real code snippet](https://github.com/agausmann/FlipFlop/blob/0087e9b579235dd4ff1a62643584c7029df58c5d/src/main.rs#L397-L401) that I had to write)

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -146,10 +146,10 @@ macro_rules! fold_array {
 /// Generate array conversion implementations for a compound array type
 macro_rules! impl_fixed_array_conversions {
     ($ArrayN:ident <$S:ident> { $($field:ident : $index:expr),+ }, $n:expr) => {
-        impl<$S> Into<[$S; $n]> for $ArrayN<$S> {
+        impl<$S> From<$ArrayN<$S>> for [$S; $n] {
             #[inline]
-            fn into(self) -> [$S; $n] {
-                match self { $ArrayN { $($field),+ } => [$($field),+] }
+            fn from(v: $ArrayN<$S>) -> Self {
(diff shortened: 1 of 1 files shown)


agausmann:
(WIP, did a more thorough search and missed some impls)

kvark:
I'm surprised Rust lets you do this, honestly. You are implementing `From`, which is a foreign trait for this crate, for stuff like `std` or `mint` types, which are also foreign for this crate. Why does it let you do this?

agausmann:
Because the signature of the impl is more than just the trait and the type you are implementing.
In `impl ForeignTrait<LocalType> for ForeignType` - the presence of the `LocalType` in the trait's generic parameter is also considered when evaluating orphan rules.

kvark:
I do wonder if this is something that changed in Rust recently (in recent years, that is). It's nice that it works though :)

kvark (approved):

PR #529, opened 2021-04-20
More Into impls converted to From

agausmann:
Followup of #528, I did a more thorough search and found some that I missed.

### DIFF
--- a/src/quaternion.rs
+++ b/src/quaternion.rs
@@ -684,7 +684,7 @@ impl<S> From<mint::Quaternion<S>> for Quaternion<S> {
 
 #[cfg(feature = "mint")]
 impl<S: Clone> From<Quaternion<S>> for mint::Quaternion<S> {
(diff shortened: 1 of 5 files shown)


agausmann:
Also appears that my auto-rustfmt changed some small things, hope that is okay

src/quaternion.rs
```diff
@@ -683,11 +683,11 @@ impl<S> From<mint::Quaternion<S>> for Quaternion<S> {
 }
 
 #[cfg(feature = "mint")]
-impl<S: Clone> Into<mint::Quaternion<S>> for Quaternion<S> {
-    fn into(self) -> mint::Quaternion<S> {
+impl<S: Clone> From<Quaternion<S>> for mint::Quaternion<S> {
+    fn from(v: Quaternion<S>) -> mint::Quaternion<S> {
```
kvark, line 687: Let's do "-> Self"?

PR #535, opened 2021-07-05
Update links in quaternions.rs

danakj:
The old link no longer goes to a valid page, and it appears to be malicious.


danakj:
I will also note here that the links don't render correctly on docs.rs: https://docs.rs/cgmath/0.18.0/cgmath/struct.Quaternion.html#performance-notes

But I am not sure what the format should be to fix it.

nstoddard:
Thanks for fixing this!

> I will also note here that the links don't render correctly on docs.rs: https://docs.rs/cgmath/0.18.0/cgmath/struct.Quaternion.html#performance-notes
> 
> But I am not sure what the format should be to fix it.

It looks like it's inserting a space between the part in [] and the part in (). Maybe try putting both parts in one line with no space in between? Not sure why that would be necessary though.

danakj:
Oh yeah, good point. Okay I've fixed the links in this file too. Thanks

kvark (approved):
Thank you!

PR #537, opened 2021-07-25
Remove NumCast from BaseNum and relax float bounds

spearman:
The first commit removes the NumCast from the BaseNum trait. This is enough to allow using non-machine scalars (e.g. fixed precision numbers) for vector/point operations:
```
#![feature(array_map)]

use cgmath::InnerSpace;
use fixed::types::I16F16 as Ftype;

macro_rules! show {
    ($e:expr) => {
        println!("{}: {:?}", stringify!($e), $e);
    };
}

fn main() {
    let p1: cgmath::Point2<Ftype> = [0, 1].map(Ftype::from_num).into();
    show!(p1);
    let p2: cgmath::Point2<Ftype> = [1, 0].map(Ftype::from_num).into();
    show!(p2);
    show!(p2 - p1);
    let v: cgmath::Vector2<Ftype> = [1, 1].map(Ftype::from_num).into();
    show!(v);
    show!(p1 + v);
    show!(v.magnitude2());
    show!(v.dot(v));
}
```
- `cast` is already parameterized by `T: NumCast`
- The `midpoint` method used cast to create a generic value of '2', but this can be replaced by `One::() + One::()`
- The only other BaseNum method using `cast` is `centroid`, to which is added an explicit `where Self::Scalar: NumCast` constraint

The second commit replaces uses of BaseFloat and Float with BaseNum and Num where possible. This might help with making finer grained trait bounds down the road (https://github.com/rustgd/cgmath/issues/496).

### DIFF
--- a/src/angle.rs
+++ b/src/angle.rs
@@ -30,7 +30,7 @@ use rand::{
 use structure::*;
 
 use approx;
-use num::BaseFloat;
+use num::{BaseFloat, BaseNum};
 
 /// An angle, in radians.
 ///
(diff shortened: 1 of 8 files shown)


kvark (approved):

kvark:
I assume this is a non-breaking change?

spearman:
> I assume this is a non-breaking change?

I think so, maybe someone can confirm.

PR #541, opened 2022-02-26
Add support for cast using bytemuck crate

maku693:
Added support for [bytemuck](https://docs.rs/bytemuck) crate.

I also applied `cargo fmt`, but I think it is out of scope, so I can remove that changes if you mind.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,7 @@ rand = { version = "0.8", features = ["small_rng"], optional = true }
 serde = { version = "1.0", features = ["serde_derive"], optional = true }
 # works only in rust toolchain up to 1.32, disabled indefinitely
 #simd = { version = "0.2", optional = true }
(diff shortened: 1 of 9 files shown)


Rua:
Good idea! I just came looking for this, and you read my mind.

kvark:
Thank you for the PR!
Does that match what any other libraries are doing?

maku693:
I think yes.

`bytemuck` [offers derive macro for that purpose](https://docs.rs/bytemuck/latest/bytemuck/derive.Pod.html), but it cannot be used for generic structs.

[glam](https://github.com/bitshifter/glam-rs/blob/e5a183a53a8037f4ce6c608ee3dd8c130d862507/src/features/impl_bytemuck.rs) does this too (but not using macros).

Rua:
Nalgebra does as well.

kvark (approved):

PR #545, opened 2022-03-13
Add From<Vector> for Point

ncihnegn:
so that Point can use unit_x() etc.


ncihnegn:
I see there is from_vec(). Why not implement From?

PR #547, opened 2022-03-26
Support infinite far planes

emilk:
Infinite far planes is a standard trick that actually improves the precision of the Z buffer.

See for instance https://thxforthefish.com/posts/reverse_z/

### DIFF
--- a/src/projection.rs
+++ b/src/projection.rs
@@ -140,6 +140,7 @@ impl<S: BaseFloat> From<PerspectiveFov<S>> for Matrix4<S> {
             persp.near
         );
 
+        let one: S = cast(1).unwrap();
         let two: S = cast(2).unwrap();
         let f = Rad::cot(persp.fovy / two);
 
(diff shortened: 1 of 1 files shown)


src/projection.rs
```diff
@@ -220,12 +230,20 @@ impl<S: BaseFloat> From<Perspective<S>> for Matrix4<S> {
 
         let c2r0 = (persp.right + persp.left) / (persp.right - persp.left);
         let c2r1 = (persp.top + persp.bottom) / (persp.top - persp.bottom);
-        let c2r2 = -(persp.far + persp.near) / (persp.far - persp.near);
+        let c2r2 = if persp.far.is_finite() {
+            (persp.far + persp.near) / (persp.near - persp.far)
```
aloucks, line 234: Should this be negative?

src/projection.rs
```diff
@@ -220,12 +230,20 @@ impl<S: BaseFloat> From<Perspective<S>> for Matrix4<S> {
...
+        } else {
+            -one
+        };
         let c2r3 = -S::one();
 
         let c3r0 = S::zero();
         let c3r1 = S::zero();
-        let c3r2 = -(two * persp.far * persp.near) / (persp.far - persp.near);
+        let c3r2 = if persp.far.is_finite() {
+            (two * persp.far * persp.near) / (persp.near - persp.far)
```
aloucks, line 243: Should this be negative?

